Issue #5, opened 2015-04-11
cannot discover devices


carlosperate:
Hi,

I am trying to run pywemo and encountering the following issue in the python terminal:
```
Python 2.7.5 (default, May 15 2013, 22:43:36) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pywemo
>>> devices = pywemo.discover_devices()
>>> print devices
[]
>>> devices = pywemo.discover_devices()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pywemo\discovery.py", line 22, in discover_devices
    device = device_from_uuid_and_location(entry.st, entry.location)
  File "pywemo\discovery.py", line 46, in device_from_uuid_and_location
    return Switch(location)
  File "pywemo\ouimeaux_device\__init__.py", line 32, in __init__
    service = Service(self, svc, base_url)
  File "pywemo\ouimeaux_device\api\service.py", line 80, in __init__
    for action in self._svc_config.get_action():
AttributeError: 'NoneType' object has no attribute 'get_action'
>>>
```
I currently have no issues discovery my Wemo switch with the ouimeaux package (although I have only tested it using these procedures http://ouimeaux.readthedocs.org/en/latest/api.html) . Any ideas what could be causing this?

carlosperate:
This commit https://github.com/balloob/pywemo/commit/e47eae6b22cdfcb90e3c235670c37f2d43eb1cbd from @tomwilkie 's fork seems fix my issue. Hopefully @tomwilkie can submit a pull request?

tomwilkie:
Need to merge and rebase, but will do that today.

Thanks

Tom

carlosperate:
Thanks Tom!

balloob:
Sorry for the delay, I have just merged it in.

PR #7, opened 2015-04-18
Fix occasional error when talking to devices

tomwilkie:
Addresses #5 - bit of a hack to just make it work.


balloob:
I'm on holiday right now for a couple of  days so unable to see whole code. Is it safe to return before the self.actions assignment?

tomwilkie:
Probably not - it does seem to work and addresses the immediate issue, but its a hack.  I'll do some more debugging today.

balloob:
Merged it as it is the best we have for now.

PR #9, opened 2015-08-02
Add power_threshold as a property

pavoni:
The power_threshold is configured into an insight switch (default is 8 - but can be changed (and autoset) in the Wemo app.

When the insight switch is turned on it looks at the consumed power - anything above this threshold is considered 'on' anything below is 'standby'.

This PR is the first step in adding an extra state into home-assistant.


balloob:
Cool, thanks!

balloob:
I'll release a new version of pywemo tonight, v0.1.1

PR #10, opened 2015-08-23
Integrate WeMo Maker code from https://github.com/iancmcc/ouimeaux, add standby state to Insight Switch

pavoni:
[~~Probably doesn't make sense to merge this until I have it working from ha, so please just treat as a heads up for now~~]

The WeMo Maker (http://www.belkin.com/uk/F7C043/p/P-F7C043/) is basically a switch and a sensor.

It switches a low voltage relay, intended to allow you to integrate with other devices. The sensor allows you to sense a TTL voltage - or if you ad a relay - anything else you like. Equipped with screw terminals, it's intended for DIY integration.

I've bought a couple to integrate with my blinds and rooflights, have only done the blinds so far. My plan is to close the blind when it's hot and the weather is sunny!

So hope you don't mind adding this integration PR.

The code was just taken from https://github.com/iancmcc/ouimeaux in line with the rest of this code; I've just done a very small amount of integration.

I've also added the code to handle diaplaying the standby mode of the Insight Switch

### DIFF
--- a/pywemo/__init__.py
+++ b/pywemo/__init__.py
@@ -7,6 +7,7 @@ from .ouimeaux_device.insight import Insight
 from .ouimeaux_device.lightswitch import LightSwitch
 from .ouimeaux_device.motion import Motion
 from .ouimeaux_device.switch import Switch
+from .ouimeaux_device.maker import Maker
 
 from .discovery import discover_devices
 from .subscribe import SubscriptionRegistry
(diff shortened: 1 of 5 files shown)


balloob:
Sounds good. Let me know when it is done.

The maker sounds very much like the insight switch, which switches power and measures usage.

pavoni:
Thanks - that's exactly what I has in mind. I'm trying to track down some sort of state problem. 

I think pywemo is incorrectly reading the device state. On the case...

pavoni:
@balloob I think ok for you to take a look now. At least it seems to work reliably for me...

balloob:
Looks good, merging it and pushing out a new release. Thanks!

PR #11, opened 2015-08-27
Fixed a crash where an empty XML string is supplied when trying to pa…

Roost:
Fixed a crash where an empty XML string is supplied when trying to parse the Wemo device description


balloob:
Awesome, thanks!

PR #13, opened 2015-10-24
Revise re-scan on error logic to match on MAC address.

pavoni:
In my setup the current re-try logic (which looks for wemo devices changing their ports etc) never finds the device.

This is because the match used for the scan never succeeded. I guess this may be router dependent.

This PR revises the retry scan logic to look for the lost device based on its MAC address.


balloob:
Looks good :+1:

PR #16, opened 2015-11-30
Bring over the bridge device from ouimeaux

kk7ds:
This brings over the relevant bits from ouimeaux to support bridge
devices (which is just the one file). It works for me to be able to
control my WeMo Link devices.


balloob:
Cool. I'll push out a new version.

balloob:
[Released as PyWemo 0.4](https://pypi.python.org/pypi/pywemo)

kk7ds:
Thanks!

PR #19, opened 2015-12-23
Update subscription code

pavoni:
This updates the subscript code to work with Python 3 - so it can be used in HA.

Closes https://github.com/balloob/pywemo/issues/17

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -98,6 +98,7 @@ class SubscriptionRegistry(object):
       self._event_thread_cond.notify()
 
   def _resubscribe(self, url, sid=None):
+    LOG.info("Wemo resubscribe for %s", url)
     headers = {'TIMEOUT': 300}
     if sid is not None:
       headers['SID'] = sid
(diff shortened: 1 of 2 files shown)


pywemo/subscribe.py
```diff
@@ -38,14 +38,13 @@ def do_NOTIFY(self):
...
-    content_len = int(self.headers.getheader('content-length', 0))
+    content_len = int(self.headers.get('content-length', 0))
     data = self.rfile.read(content_len)
-
     if device is None:
       LOG.error('Got event for unregistered device %s', sender_ip)
     else:
       # trim garbage from end, if any
-      data = data.split("\n\n")[0]
+      data = data.decode().split("\n\n")[0]
```
balloob: Could you provide an encoding here instead of using 'default' encoding?

pywemo/subscribe.py
```diff
@@ -57,7 +56,7 @@ def do_NOTIFY(self):
     self.send_header('Content-Length', len(SUCCESS))
     self.send_header('Connection', 'close')
     self.end_headers()
-    self.wfile.write(SUCCESS)
+    self.wfile.write(SUCCESS.encode())
```
balloob: Could you provide an encoding here instead of using 'default' encoding?

PR #29, opened 2016-02-09
Allow connecting to wemo devices directly

kk7ds:
This patch addresses two issues for me:
1. I'd like to be able to talk to wemos directly over a VPN by IP
2. The discovery mechanism is very unreliable at times

This patch allows this behavior by letting a user specify a device
manually by the URL needed to reach it. In order to make this work,
we need to implement the rediscovery procedure differently because
we can't expect to be able to find it via SSDP.

If mac is set to None, then we probe for the port the wemo is currently
using, based on the list of seen-in-the-wild ports we know they can
be on.

I've been using a variant of this in a local fork of pywemo for a
while and have found it to be much more reliable than depending on
discovery for initial setup and rediscovery.

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -13,6 +13,29 @@ from .api.xsd import device as deviceParser
 
 log = logging.getLogger(__name__)
 
+PROBE_PORTS = (45152, 49153, 49154)
+
+
+def probe_wemo(host):
+    """Probe a host for the current port.
+
+    This probes a host for known-to-be-possible ports and
+    returns the one currently in use. If no port is discovered
+    then it returns None.
+    """
+    for port in PROBE_PORTS:
+        try:
+            r = requests.get('http://%s:%i/setup.xml' % (host, port),
+                             timeout=10)
+            if 'WeMo' in r.text:
+              return port
+        except requests.exceptions.Timeout:
+            # Timeout means it's not worth trying other ports
+            return None
+        except requests.exceptions.ConnectionError:
+            pass
+    return None
+
 
 class UnknownService(Exception): pass
 
@@ -36,7 +59,7 @@ class Device(object):
             self.services[svcname] = service
             setattr(self, svcname, service)
 
-    def reconnect_with_device(self):
+    def _reconnect_with_device_by_discovery(self):
         """
         Wemos tend to change their port number from time to time.
         Whenever requests throws an error, we will try to find the device again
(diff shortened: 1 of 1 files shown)


kk7ds:
Hi,

I'd like to add static wemo support to HA and need something like this available. I've been using a local fork of this approach with a lot of success for a couple of months.

What do you think?

kk7ds:
By the way, you can test this locally with something like this:

```
import logging
import pywemo

s = pywemo.Switch('http://192.168.50.183:49153/setup.xml', None)
s.reconnect_with_device()
```

pavoni:
Certainly agree discovery is shaky. How do you initially find the devices (and at least get the ip addresses)? How safe is the Port range? Guess I should check / trace mine...

kk7ds:
My local wemos are quarantined onto a separate subnet that I isolate from the internet (and the other subnets). Thus, I control that one pretty tightly, which includes the addresses of the devices. Since they never change address, there's really no need to have the discovery process going. Plus, being across subnets, the discovery won't work (or requires some roundabout hacks to make it work).

The other use is for some wemos we have in a vacation home that I can access over a VPN. I don't want to do discovery over that tunnel of course, so I'd like to be able to point at them directly. They don't ever change address, but if they did, I can collect the addresses from my router's DHCP lease log.

kk7ds:
This is what I was going to propose for HA:

http://pastebin.com/wVzPhByx

kk7ds:
Oh, and on the port range thing... After running for a few months, I've seen them cycle mostly between 49152 and 49153. I saw 49154 one time after a semi-weird restart during a firmware restart process. This is across wall switches, the original plug switch, insights, and one wemo bridge. None of them ever jumped outside that range, so I think it's fairly safe.

Since the probing goes really quick, we could expand it to 49150-49160 or something if you want. I've just never seen it be necessary.

pavoni:
I think this is a good idea. Rediscovery is shaky even on a network where it does work. It also seems that changing ports is much more common than changing ip.

So I think what would make sense is to combine the two approaches, so if we lose contact we try your probe first - and fall back to discovery if we need to. 

Would you be OK to change to do this? I'll double check my port numbers - but what you say makes sense.

kk7ds:
Sure, no problem. I thought about doing this first, since usually a port-rescan goes really fast (unless the device is gone, and then it's just the timeout). I just didn't want to overly alter the discovery mechanism for those that use it.

I'll cook up a change later in the morning.

Thanks!

kk7ds:
@pavoni Did you see that I updated the commit? Thoughts?

Thanks!

pavoni:
I didn't - will take a look - thanks.

pavoni:
Looks good - thanks.

pavoni:
Released as 0.3.10. Thanks.

PR #30, opened 2016-02-16
Fixes for WeMo led devices

jaharkes:
- autodiscovery for the bridge device
- make sure cached data about lights and groups gets refreshed


pavoni:
Looks good - thanks!

PR #32, opened 2016-02-17
Probe for a port to listen on for subscription callbacks

kk7ds:
This probes for an available port, starting at the current default of
8989 for an available port to listen on. This avoids failure if we
have a conflict for the current statically-configured port.

In the extremely unlikely event that we can't bind to anything, this
raises subscribe.SubscriptionRegistryFailed.

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -18,8 +18,11 @@ import requests
 LOG = logging.getLogger(__name__)
 NS = "{urn:schemas-upnp-org:event-1-0}"
 SUCCESS = '<html><body><h1>200 OK</h1></body></html>'
-PORT = 8989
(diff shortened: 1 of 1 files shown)


kk7ds:
I tested this, not not against a real wemo because of $reasons.

I assume you can give it a whirl for real and make sure it's good. If not, I'll do it later and report back.

pavoni:
Looks good - thanks Dan. Should be able to check on some physical devices this eve (if you haven't already). 

Also added you to the Repo.

kk7ds:
Thanks @pavoni!

I tested this locally and made one change as a result, which was just a python2 thing of needing to handle socket.error as well as OSError when detecting the conflict.

Issue #33, opened 2016-02-22
wemo link improvements


jaharkes:
Although light support is mostly working, some things were bothering me. For instance when setting the state to turn a light off, the dim value has to be set to 0 as well otherwise the bulb turns off for a fraction of a second and then goes back to the dimmed value (mostly noticeable in peripheral vision).

Also wemo supports fading (transitions) and when paired with some (non-wemo) bulbs and/or light strips supports things like xy color values and color temperatures. It turns out a lot of this has been figured out months ago and posted in the ouimeaux bugtracker (iancmcc/ouimeaux#51).

The important bits of information are in the following gists,
- List of light capabilities with descriptions https://gist.github.com/hardillb/ec88e86597d65584b1ba taken from an Osram bulb paired with a WeMo bridge.
- node.js code to control lights https://gist.github.com/hardillb/1279241bb886ee28c05b which is not setting multiple capabilities in a single request the way the current ouimeaux/pywemo code is.

jaharkes:
I picked up an Osram/Sylvania full color, tunable white bulb last weekend and have been poking at it with pywemo.

First of all, the bulb turns out to be not "WeMo Certified", but does pair with the WeMo Link Zigbee bridge device. The android app does not handle things well though, it sits endlessly spinning waiting for the pairing to complete and becomes unresponsive. After killing and restarting the app the bulb never shows in the main ui, only when you dig deep into the menus (settings/about/firmware) the firmware version is shown.

On the pywemo side life is a lot better. Immediately after pairing in the android app begins we can query the bridge and see the new bulb show up with a generic name, "Color Light 03", I assume this is partly defined by the bridge since this was the third light I have paired. The default on/off and brightness controls work but turning off performs a 1-2 second fade-out instead of a fast off, setting brightness to 0 with a 0 transition time does make the bulb turn off immediately.

I made some modifications and have been able to change both color temperature and rgb color. Both of these are very similar to how the Philips Hue bulbs are set by home-assistant, this probably follows from the fact that there is a Zigbee LightLink standard which different vendors try to comply with. Temperature is defined in mireds and can be set in the range [151,555] which corresponds to about 6600K to 1800K. Color is set as color X and Y values in the range of [0,65535].

The main question right now is probably what a usable API for this would look like. Right now I'm mostly throwing raw values back and forth which isn't too user friendly. I'll clean up what I have and push, hopefully in a day or two.

PR #34, opened 2016-03-13
WeMo Link add support for tunable white and RGBW lights

jaharkes:
This pretty much rewrites the interface how we talk to zigbee devices through a wemo bridge, so this is really a request for comments.
- Using elementtree to format the setdevicestatus xml request.
- Added support for colortemperature and colorcontrol, see #33  (tested with Osram Lightify RGBW)
- Added per-light/group instances which reduce the number of arguments we need to pass when controlling the lights.
- On the bridge, index devices based on unique device identifier, instead of the user-configured name.

### DIFF
--- a/pywemo/color.py
+++ b/pywemo/color.py
@@ -1,0 +1,55 @@
+# Define usable ranges as bulbs either ignore or behave unexpectedly
+# when it is sent a value is outside of the range.
+TEMPERATURE_PROFILES = dict((model, temp) for models, temp in (
+    # Lightify RGBW, 1900-6500K
+    (["LIGHTIFY A19 RGBW"], (151, 555)),
+) for model in models)
+
+COLOR_PROFILES = dict((model, gamut) for models, gamut in (
+    # Lightify RGBW, 1900-6500K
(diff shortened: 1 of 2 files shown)


pavoni:
The code looks very good. I don't have any Lights - so I can't test myself. 

Your PR says you're looking for comments - but the code looks pretty finished!

Are you happy for me to merge / release a new version - or are you still testing.?

I'll give you access to the repo...

Thanks for the contribution.

jaharkes:
Comments because it really changes around how we talk to linked ZigBee devices and drops any backward compatibility with previous versions. OTOH we didn't really have lights working until recently.

Until now I've only been working with this from a interactive Python session directly using the pywemo APIs, I just started modifying the home assistant code to use the new functionality. Cleans up the HA side a bit, and I've already found that the sleepfader control support probably should get folded into the turn_off() function.

pavoni:
Sounds good - let me know if I can help - and happy to merge when you're ready.

We could build an interim version to make it easier for third parties to test with HA if that helps at some point.

jaharkes:
It is a good thing HA has a strong binding to a particular version, so any incompatible release won't accidentally break existing installations.

But yes, being able to test this will be useful.

PR #35, opened 2016-03-15
Fix crash when SSDP response has empty device description.

jaharkes:
Bug was reported by @liocer on the home-assistant gitter. This is probably the same problem as in #11, but hitting in a different place of the discovery code.


pavoni:
Thanks 0.1.13 now pushed.

PR #36, opened 2016-03-16
Revise probe wemo

pavoni:
Turns out that the wrong port can timeout, so keep searching.

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -30,8 +30,7 @@ def probe_wemo(host):
             if 'WeMo' in r.text:
               return port
         except requests.exceptions.Timeout:
-            # Timeout means it's not worth trying other ports
-            return None
+            continue
         except requests.exceptions.ConnectionError:
             pass
     return None
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup, find_packages
 
 setup(name='pywemo',
-      version='0.3.13',
+      version='0.3.14',
       description='Access WeMo switches using their SOAP API',
       url='http://github.com/pavoni/pywemo',
       author='Greg Dowling',


kk7ds:
How can that happen?

If the wemo is up, then it will return a RST on any port other than the open one, which will immediately give us a connection error. If it doesn't, then it should mean that it's not up at all (or not at the same address). Have you experienced this with some device?

pavoni:
Yes - currently happening in my network!

pavoni:
```
>>> pywemo.ouimeaux_device.probe_wemo('192.168.1.193')
Timeout on port 49153
Success on port 49154
49154
```

kk7ds:
I really don't see how that's possible. If you do something like this:

```
telnet 192.168.1.20 49152
```

Using the IP of the wemo of course, do you get a connection refused or a timeout? If the latter, are you sure it's not a packet loss problem? Meaning, a busy WiFi network preventing the wemo from responding quickly or something?

The reason I'm obsessing over this is just that incurring a 10-20 second timeout per port when re-probing makes things a lot less useful because if you go to hit the switch and the wemo has moved ports, the re-probe should be able to locate the proper port in a short period of time such that it's just a slight delay between flipping the switch and the wemo responding, instead of what could be close to a minute :(

kk7ds:
Huh, okay, well, that's super disappointing (and I've never seen it happen here).

pavoni:
Of course may be a network issue. But discovery finds it fine. This is the code from above

```
def probe_wemo(host):
    """Probe a host for the current port.

    This probes a host for known-to-be-possible ports and
    returns the one currently in use. If no port is discovered
    then it returns None.
    """
    for port in PROBE_PORTS:
        try:
            r = requests.get('http://%s:%i/setup.xml' % (host, port),
                             timeout=10)
            if 'WeMo' in r.text:
              log.error('Success on port %s', port)
              return port
        except requests.exceptions.Timeout:
            log.error('Timeout on port %s', port)
            continue
        except requests.exceptions.ConnectionError:
            pass
    return None
```

Was switching to use static addresses - and stumbled into the issue.

pavoni:
Telnet claims to connect and then hangs

```
❯ telnet 192.168.1.192 49152                                                  ⏎
Trying 192.168.1.192...
Connected to 192.168.1.192.
Escape character is '^]'.
```

On port 4153 it connects and then terminates.

```
❯ telnet 192.168.1.192 49153                                                  ⏎
Trying 192.168.1.192...
Connected to 192.168.1.192.
Escape character is '^]'.
HTTP/0.0 400 Bad Request
CONTENT-LENGTH: 50
CONTENT-TYPE: text/html

<html><body><h1>400 Bad Request</h1></body></html>Connection closed by foreign host.

[greg@Gregs-MacBook-Pro] ~/ha-python/pywemo (master) ⚡
❯ telnet 192.168.1.192 49154                                                  ⏎
Trying 192.168.1.192...
telnet: connect to address 192.168.1.192:
```

on 49154 it fails to connect.

```
❯ telnet 192.168.1.192 49154                                                  ⏎
Trying 192.168.1.192...
telnet: connect to address 192.168.1.192: Connection refused
telnet: Unable to connect to remote host
```

kk7ds:
Just when I thought wemos couldn't be more disappointing... Just confirmed that none of mine do that. Do all of yours do it or is it limited to one class of device?

I really wanted to not have to run a forked pywemo, but I guess I'm just going to have to do it. If my devices started filtering ports like that (if that's what is going on) such that I couldn't work around their port hopping like this, I'd just dump them altogether :/

Anyway, thanks for the extra info.

jaharkes:
Maybe we could try multiple ports in parallel.

pavoni:
It's just one - could reboot - but my general feeling is it should work at least as well as discovery.

We could add a config parameter to HA.

Parallel sounds like a cool idea.

kk7ds:
@pavoni Wow, that is super weird behavior of accepting the connection but not doing anything. That's pretty broken :(

jaharkes:
Seems like a perfect case for 'concurrent.futures'
https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor-example

pavoni:
Just realized the telnet session was to the wrong wemo - but the 192.168.1.193 telnet does exactly the same.

pavoni:
Or we could try the `right` way first - and fall back to the wrong way.

kk7ds:
Based on what you're seeing, we're hitting a request timeout, not a socket timeout, which is what I was going for. A socket timeout means the thing didn't respond at all. A request timeout will occur when you issue the HTTP request and get no response (like your telnet). Not sure if we can identify which it was from the requests exception.

We could also try a few times, upping the timeout each time. First at 1 second, then 5, then 10. That would mean if we do hit that stupid situation we'll be 2-3 seconds per port probe, instead of 30, but still be tolerant of high-latency connections.

kk7ds:
Ah, I got it. We can catch ConnectTimeout and ReadTimeout separately.. Bail on the former and pass on the latter. I'll write a patch for you to test.

pavoni:
Sorry turn on brain before commenting!

pavoni:
Thanks @kk7ds!

PR #39, opened 2016-03-19
Reduce chatter when turning lights on.

jaharkes:
Accept extra arguments to the turn_on() call to reduce the number of requests that need to be sent from the bridge to the bulb from 4 to 2 when turning on a bulb with a color or temperature value.


jaharkes:
I think I might have a cleaner way of doing this by pushing commands onto a queue.

pavoni:
Released as 0.3.19. Thanks.

PR #41, opened 2016-03-29
SSDP discover fixed in ssdp.py

gitaaronc:
Lets try this again as I am a git NOOB.


gitaaronc:
I found what seems to be a better solution. I will make a new pull request.

PR #42, opened 2016-03-30
Bind to ip address of host name in ssdp::scan function

gitaaronc:
This fixes my Windows 10 issue but messes up linux, as my linux machine hostname is mapped to 127.0.0.1 where the Windows machine is resolving to the IP address of the ethernet card.
My solution on linux is to modify the /etc/hosts file so the hostname points to the IP address you wish to use for the source address. The other option would be to redesign and pass in the source IP address on the command line or via a config file.

Thoughts?


gitaaronc:
jaharkes fix is better.
https://github.com/pavoni/pywemo/pull/43

PR #43, opened 2016-03-30
Possible fix for discovery on Windows

jaharkes:
Bind sockets to interface addresses when sending discovery requests in an attempt to fix sending discovery requests on Windows systems.


jaharkes:
Will need some confirmation that this actually fixes the problem, I don't have any Windows systems to test it on.

It looks like the ssdp discovery code here is identical to what is used in balloob/netdisco so if this fix works it should probably be propagated there as well.

And maybe pywemo should then use the discovery functionality provided by netdisco instead of duplicating the code.

gitaaronc:
I'll give it a try when I get home I a few hours although this isn't much different then my original approach and includes an additional dependency.

jaharkes:
Of course it isn't much different from your original approach, because that was a way that worked for you but broke other systems. The additional dependency is needed because it is the cross-platform way to get interface addresses on Windows, Linux and OS X.

This version also parallelizes waiting for the discovery responses, so it only waits for one timeout period instead of a full timeout for each visible IPv4 network.

gitaaronc:
Your changes work for me on linux and Windows 10.

pavoni:
Works for me - merged. Thanks for the contribution.

Getting a 500 on uploading to pypi - so will try again later - and confirm the version number.

Will you do the HA PR or shall I?

I like your idea of using netdisco - that way you'll be maintaining one piece of code :-)

jaharkes:
I can handle the HA PR, have the same fix queued up as balloob/netdisco#20, the only difference I think is that code sets the multicast TTL to 1 hop so maybe the HA PR can update both at the same time.

It is interesting to see the similarities, I am guessing it started as balloob/pywemo and then was generalized to netdisco. Also a little surprised how different the code seems to be from home-assistant, no copyright blurbs on files, no flake8/pylint/pytest/tox setup.

pavoni:
Now released as 0.3.23. Will leave HA to you (thanks).

We could make a call to properly take wemo through flake8/pylint - and add some tests.

I took some of the files through flake8/pylint when I took over - but stopped when it would have led to divergence from ouimeaux without clear benefit. Up for it if you'd like to.

I think the revised discovery was one of the main things baloob changed - to simplify the environment and dependencies.

PR #49, opened 2016-08-10
Change encoding to "utf-8".

mickeprag:
Using the encoding "unicode" raises the error:
LookupError: unknown encoding: unicode


pavoni:
Thanks. Will release a new library today.

pavoni:
Now released as 0.4.4

PR #53, opened 2016-08-17
Update bridge.py

cbrunet:
Python 3.5. When I tried to turn on a light, I got the following error (from ElementTree):
TypeError: string argument expected, got 'bytes'

My edit seems to solve the problem. However, you should check it still works in Python 2.

### DIFF
--- a/pywemo/ouimeaux_device/bridge.py
+++ b/pywemo/ouimeaux_device/bridge.py
@@ -76,10 +76,10 @@ class Bridge(Device):
         et.SubElement(req, 'CapabilityID').text = ','.join(capids)
         et.SubElement(req, 'CapabilityValue').text = ','.join(values)
 
-        buf = six.StringIO()
+        buf = six.BytesIO()
         et.ElementTree(req).write(buf, encoding='utf-8',
                                   xml_declaration=True)
-        sendState = html_escape(buf.getvalue(), quote=True)
+        sendState = html_escape(str(buf.getvalue()), quote=True)
         return self.bridge.SetDeviceStatus(DeviceStatusList=sendState)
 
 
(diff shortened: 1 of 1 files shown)


pavoni:
Thanks for the contribution, I can't check this myself since I don't have a bridge @jaharkes @kk7ds are you comfortable?

pywemo/ouimeaux_device/bridge.py
```diff
@@ -76,10 +76,10 @@ def bridge_setdevicestatus(self, isgroup, deviceid, capids, values):
         et.SubElement(req, 'CapabilityID').text = ','.join(capids)
         et.SubElement(req, 'CapabilityValue').text = ','.join(values)
 
-        buf = six.StringIO()
+        buf = six.BytesIO()
         et.ElementTree(req).write(buf, encoding='utf-8',
                                   xml_declaration=True)
-        sendState = html_escape(buf.getvalue(), quote=True)
+        sendState = html_escape(str(buf.getvalue()), quote=True)
```
kk7ds: Just change this to:

```
buf.getvalue().decode()
```

Instead of the str() cast.

kk7ds: This is why:

```
>>> b = six.BytesIO()
>>> b.write(b'foo')
3
>>> b.getvalue()
b'foo'
>>> b.getvalue().decode()
'foo'
>>> str(b.getvalue())
"b'foo'"
```

pavoni:
Thanks @cbrunet and @kk7ds - now released as 0.4.6

@cbrunet assume you did this to use with HA - do you want to do the PR to upgrade the HA version - or would you like me to.
- Greg

cbrunet:
I wasn't aware of home assistant. Yes, you can do the pr for ha.

PR #55, opened 2016-12-03
Handle cases where the bridge doesn't know the brightness

mjg59:
If my bridge is power cycled then it has no idea what brightness the bulbs
are at, and appears to default to not providing that information on query.
If a client then tries to turn on the bulb but doesn't provide a level, we
get a KeyError. In the absence of any other information, assume that we want
the lights on full in that case.

### DIFF
--- a/pywemo/ouimeaux_device/bridge.py
+++ b/pywemo/ouimeaux_device/bridge.py
@@ -215,8 +215,11 @@ class Light(LinkedDevice):
             # - After turning off a bulb with sleepfader, it fails to turn back
             #   on unless the brightness is re-set with levelcontrol.
             self.get_state(force_update=force_update)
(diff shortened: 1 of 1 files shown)


pavoni:
Looks good to me - although I don't have a bridge. 

@jaharkes does this make sense to you?

jaharkes:
Looks good to me, although I probably would have used state.get("level", 255). But this does the same thing.

pavoni:
Certainly more pythonic! 

@mjg59 if you can make the change @jaharkes suggests I'll merge and release a new version.

mjg59:
Updated, thanks for the feedback!

pavoni:
Thanks for the contribution, now released as `0.4.8`

PR #59, opened 2017-01-18
Coffee maker

stu-gott:
Add support for Wemo Coffee Makers

### DIFF
--- a/pywemo/__init__.py
+++ b/pywemo/__init__.py
@@ -8,6 +8,7 @@ from .ouimeaux_device.lightswitch import LightSwitch
 from .ouimeaux_device.motion import Motion
 from .ouimeaux_device.switch import Switch
 from .ouimeaux_device.maker import Maker
+from .ouimeaux_device.coffeemaker import CoffeeMaker
 from .ouimeaux_device.bridge import Bridge
 
 from .discovery import discover_devices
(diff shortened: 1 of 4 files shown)


pavoni:
@stu-gott thanks for the contribution.

Your host change looks good - will double check on my set-up - but your suggestion makes sense.

There looks to be some subscription code commented out - and a `subscription_callback` function that isn't used at this point.

I suspect you had in mind to do what I never got to - which is to use the callback data to update the device state.

Very happy for you to do this (although it would need careful testing) - also happy to merge a simplified version that works like the rest of the library.

The library is mostly used by home_assistant at this point - and HA just calls for a state update when it receives the callback. It's not perfect - but better that the polling it did before!

Let me know what you'd like to do. Do you plan to use this directly - or in HA?

Thanks again - Greg

stu-gott:
this is for HA. I will commit my patch for that to my fork when I get a
chance. I saw where you were using update() and that's what
subscription_callback is for here. the commented code was an example of how
to use that function. but since I closed the loop and have a patch for HA
too that's pretty much irrelevant at this point.

stu-gott:
https://github.com/stu-gott/home-assistant/tree/wemo-coffeemaker

This is the tree with the matching HA changes in place.

pavoni:
Thanks - now released as `0.4.10`.

Thanks for the contribution.

PR #60, opened 2017-01-24
Properly update internal state for async updates.

stu-gott:
There was a bug in the async_callback where device._state was not
updated unless get_state(force_update=True) was used.


pavoni:
Now released as `0.4.11`

Issue #64, opened 2017-02-14
Troubleshooting insightparams subscriptions for Home Assistant


deftdawg:
I'm not really a python guy, but I'm trying to troubleshoot why Home Assistant's Wemo switch component does not seem to be receiving insightparam update events (https://github.com/home-assistant/home-assistant/issues/3506)... 

It is able to get insightparams when the device is first discovered when home assistant starts. 

I was able to validate that the Wemo Insight is working by testing against the npm wemo-client package with the following program (updates come about 1 per second):
```
var Wemo = require('wemo-client');
var wemo = new Wemo();
const util = require('util')

wemo.discover(function(deviceInfo) {
  console.log('Wemo Device Found: %j', deviceInfo);
  // Get the client for the found device
  var client = wemo.client(deviceInfo);

  // Handle BinaryState events
  client.on('insightParams', function (binaryState, instantPower, data) {
    console.log('insightParams changed to: %s %s', instantPower, util.inspect(data, {showHidden: false, depth: null}));
  });
});
```

However I've been unable to find a simple subscriptionRegistry example in Python I can use to validate that pywemo is capturing the insightParams events.   This is as far as I got and I'm obviously missing something because I'm not getting any log messages after it says `INFO:pywemo.subscribe:Listening on port 8990` (I probably need an event handler, but not sure how to add it)

```
#!/usr/bin/python3

import sys
import pywemo
import logging
devices = pywemo.discover_devices()
print(devices)
logging.basicConfig(level=logging.DEBUG)
SUBSCRIPTION_REGISTRY = pywemo.SubscriptionRegistry()
SUBSCRIPTION_REGISTRY.start()
```

Hoping someone can point me in the right direction.  Thanks.

pavoni:
This sounds a bit like this issue
https://github.com/home-assistant/home-assistant/issues/3506

I suspect the the wemo device only triggers a callback when something `significant` happens - such as a device turning on / off - or power useage crossing the threshold.

I have a coffee machine that I detect a power useage change when the thermostat triggers (and the power useage crosses the threshold) - but I don't think it sends other messages based on power useage.

Well worth investigating more though!

To get your subscription example to work you need to do something like this:-

https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/switch/wemo.py#L60

Also please note that at this point you have to call the update function to fetch state after you get a callback, we don't yet extract state from the callback message.

deftdawg:
[...]

PR #65, opened 2017-02-14
Fix for #64

deftdawg:
Fix for #64

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -111,7 +111,8 @@ class SubscriptionRegistry(object):
       })
     try:
       url = device.basicevent.eventSubURL
-      if hasattr(device, 'insight'): # HACK: Insight updates instead of basicevent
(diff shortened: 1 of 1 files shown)


pavoni:
This looks good - thanks for all the detective work.

Give me a chance to test against my devices- and then I'll merge.

I also wonder if we add the same test for the `bridge` (see https://github.com/timonreinhard/wemo-client/blob/master/client.js#L46) we might get realtime events for bridges. I don't have a bridge but  @jaharkes might be interested.

jaharkes:
That is definitely interesting because right now I don't think it picks up state changes very well and I was blaming any discrepancies on the hardware but it could very well be caused by how we talk to it.

My bridge has been having some trouble pulling the latest version of the firmware, once I have it current I'll look into this.

pavoni:
Released as part of `0.4.13` thanks for your detective work and contribution.

PR #69, opened 2017-03-25
Ability to use subscription to update insight switch.

pavoni:
This steals the idea from @stu-gott to use the subscription update to set the device state for Insight switches - rather than refetching.

 This is a bigger win now we're getting updates for Insight devices every second.

@stu-gott At some point I'll finish the refactor - and add this to other devices. Could you let me know what `_type` is logged for the `CoffeeMaker`? My plan is to pass unknown subscription types up the object hierarchy (eg so the base `Switch` can handle the basic on off updates).

I'll merge and release - so this can easily be tested in HA - but happy to fix any concerns or issues.

### DIFF
--- a/pywemo/ouimeaux_device/coffeemaker.py
+++ b/pywemo/ouimeaux_device/coffeemaker.py
@@ -72,7 +72,7 @@ class CoffeeMaker(Switch):
         self._attributes = attributeXmlToDict(resp)
         self._state = self.mode
 
-    def subscription_callback(self, xmlBlob):
+    def subscription_update(self, _type, xmlBlob):
         """
         Handle reports from device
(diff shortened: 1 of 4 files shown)


stu-gott:
```>>> 2017-03-25 16:40:10,285 - pywemo.subscribe - INFO - Received event from <WeMo CoffeeMaker "CoffeeMaker">(192.168.1.30) - attributeList <attribute><name>ModeTime</name><value>0</value><prevalue>1</prevalue><ts>1490474409</ts></attribute><attribute><name>Mode</name><value>3</value><prevalue>1</prevalue><ts>1490474409</ts></attribute>```
device _type: attributeList

PR #71, opened 2017-03-31
Fix compatability with Sylvania tunable white bulbs

emunsing:
It seems that Osram/Sylvania Lightify tunable white bulbs report their status as float strings, which led to a typecasting issue when casting directly to int. This has now been tested to work with the Sylvania Lightify A19 Tunable White 60W equivalent bulb.

### DIFF
--- a/pywemo/ouimeaux_device/bridge.py
+++ b/pywemo/ouimeaux_device/bridge.py
@@ -105,7 +105,7 @@ class LinkedDevice(object):
             if not value:
                 value = None
             elif ':' in value:
-                value = tuple(int(v) for v in value.split(':'))
+                value = tuple(int(float(v)) for v in value.split(':'))
             else:
                 value = int(value)
             status[capability] = value
--- a/pywemo/ouimeaux_device/bridge.py
+++ b/pywemo/ouimeaux_device/bridge.py
@@ -105,7 +105,7 @@ class LinkedDevice(object):
             if not value:
                 value = None
             elif ':' in value:
-                value = tuple(int(float(v)) for v in value.split(':'))
+                value = tuple(int(round(float(v))) for v in value.split(':'))
             else:
                 value = int(value)
             status[capability] = value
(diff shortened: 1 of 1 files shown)


pywemo/ouimeaux_device/bridge.py
```diff
@@ -105,7 +105,7 @@ def _update_state(self, status):
             if not value:
                 value = None
             elif ':' in value:
-                value = tuple(int(v) for v in value.split(':'))
+                value = tuple(int(float(v)) for v in value.split(':'))
```
pavoni: Probably best to do `int(round(float(v)))`  which will round rather than truncate - and work on Python 2 & 3.

pywemo/ouimeaux_device/bridge.py
```diff
@@ -105,7 +105,7 @@ def _update_state(self, status):
...
             else:
                 value = int(value)
```
pavoni: Could you also do `int(round(float(v))) `. Probably not needed for your devices - but could help with some other device!

pavoni (changes requested):
Thanks - looks good - couple of minor suggestions, then happy to merge and release. 

Assuming you're using this with HA - would you like to do the PR - or should I?

emunsing:
Great, good catch- I like the update, and have pushed a new commit that addresses this. Let me know if there's anything else you catch! I'm new to pull requests, so let me know what's needed to wrap this up :)

pywemo/ouimeaux_device/bridge.py
pavoni: @emunsing Don't know if you saw this - could you add the same code in line 110?

pavoni:
@emunsing One small suggestion that I think you missed (doing the same cast in the `else`).

Then I'll to merge and release. Thanks for the contribution.

emunsing:
Great, thanks for the catch- updated!

pavoni (approved):

pavoni:
Released as 0.4.16. Thanks.

Would you like to do the HA pull request - or shalI?

emunsing:
I'm not actually working with the the Home Assistant system, if that's what you're referring to by HA- go ahead and do the pull request!

PR #72, opened 2017-04-09
Use common format for CoffeeMaker and Insight

stu-gott:
update_insight_params is not a method that Switch has. I assume this should have been subscription_update.

Tested on HA 0.42.2


pavoni:
Apologies - and thanks for the PR - now released as 0.4.17. I'll sort the HA PR.

Issue #76, opened 2017-06-22
devices with multiple switches on different ports not handled


ausil:
the probe_wemo function seems to have an assumption that a ip can only ever have a single devices and listening port. 

I have some devices that has 8 switched ports it is emulating a wemo device. it is listening on ports 49153-49159. homeassistant is using pywemo in part of the discovery. currently it only finds a single switch for each device. I think I almost have a patch ready to be submitted. but wanted to see if you had ideas on how it should be done

pavoni:
I'm going to close this for now - since this isn't an official wemo device.

Open to a new PR...

PR #78, opened 2017-06-22
attempt to handle probing devices that offer multiple switches

ausil:
devices such as https://www.digital-loggers.com/pro.html emulate a
wemo device. They offer multiple ports by having each switch port
use a different tcp port. added an optional argument to probe_wemo
function to pass in the description and have an extra check. In my
testing it worked to detect my devices and get the tcp ports correct

Signed-off-by: Dennis Gilmore <dennis@ausil.us>

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -14,7 +14,8 @@ from .api.xsd import device as deviceParser
 log = logging.getLogger(__name__)
 
 # Start with the most commonly used port
-PROBE_PORTS = (49153, 49152, 49154, 49155, 49156, 49157, 49158, 49159)
+PROBE_PORTS = (49153, 49152, 49154, 49155, 49156, 49157, 49158, 49159,
+        49160)
 
 
 def probe_wemo(host, description=None):
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -14,10 +14,10 @@ from .api.xsd import device as deviceParser
 log = logging.getLogger(__name__)
 
 # Start with the most commonly used port
-PROBE_PORTS = (49153, 49152, 49154)
+PROBE_PORTS = (49153, 49152, 49154, 49155, 49156, 49157, 49158, 49159)
 
 
-def probe_wemo(host):
+def probe_wemo(host, description=None):
     """Probe a host for the current port.
 
     This probes a host for known-to-be-possible ports and
@@ -28,6 +28,17 @@ def probe_wemo(host):
         try:
             r = requests.get('http://%s:%i/setup.xml' % (host, port),
                              timeout=10)
+            if description is not None:
+                # for devices that listen on multiple ports try and 
+                # match the device description.
+                # there is a possiblity this will get things wrong.
+                # in order to cover all the cases we would need to
+                # parse the xml and match on an exact filed content,
+                # having ports named the same would go wrong also.
+                if description not in r.text:
+                    log.debug('port %s does not match our device %s',
+                             port, description)
+                    continue
             if ('WeMo' in r.text) or ('Belkin' in r.text):
                 return port
         except requests.exceptions.ConnectTimeout:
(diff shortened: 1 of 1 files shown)


ausil:
fixes #76

pavoni:
I'm away from my laptop and wemo for 10 days -but will take a look when I'm back.

pavoni:
Looks good, will double check it against my wemo devices.

Only concern is the time taken to probe extra ports. Start-up is already slower that it would ideally be - and so adding extra ports might make it worse. 

Not sure what we can do...

pavoni:
Had a thought - https://github.com/home-assistant/home-assistant/pull/8307

ausil:
the probe should not be any slower for belkin devices, the probing of ports stops once it matches on a port. I have 4 insight switches on a different subnet that in hass are statically configured and two of the digital loggers units I have found the performance to be acceptable.

ausil:
```
time python3 -c "import pywemo; devices = pywemo.discover_devices(); print(devices)"
Found malformed XML at http://10.46.1.44:9080: status=ok
[<WeMo Switch "athosian">, <WeMo Switch "unused">, <WeMo Switch "Fedora MAC">, <WeMo Switch "spare monitor">, <WeMo Switch "jup">, <WeMo Switch "wandboard02">, <WeMo Switch "m3d">, <WeMo Switch "bbb02">, <WeMo Switch "anubis">, <WeMo Switch "28_Asus_Monitor">, <WeMo Switch "22_Asus_Monitor">, <WeMo Switch "Speakers">, <WeMo Switch "KVM switch">, <WeMo Switch "cubietruck01">, <WeMo Switch "hathor">, <WeMo Switch "unused_pwr02_port8">]

real	0m13.911s
user	0m0.897s
sys	0m0.070s
```

pavoni:
After more thought - I don't like the idea of adding the device name to probe-wemo.

In the case of this device - you don't want to probe it anyway, either initially - or when something goes wrong.

I think a cleaner approach would be to use the discovery function directly.
````
>> url = 'http://%s:%i/setup.xml' % (address, port)
>> device = pywemo.discovery.device_from_description(url, None)
````
And then we'd need to find a way of disabling re-probes on failures.

In terms of Home Assistant - it may be better to add this a a new switch platform. This would mean you don't need to worry about the wemo discovery code and can just use the code above directly - and you can just add every switch at an IP address - and you don't need to disable the subscription code and turn polling off in the wemo code.

ausil:
you would still have to probe it, in the switch configuration you can choose which ports to expose via upnp, so you could have anywhere from 1-8 ports to control.

pavoni:
How are you discovering the device? Via upnp discovery - or via static address?

ausil:
I am using upnp discovery, I have enabled in both of my units all 8 ports to be exposed, though I probably should disable the ones I'm not using. At the least I need to test what it looks like. I'm not sure if the ports are tied to the switch or if it just starts at the bottom and counts up

pavoni:
So do the discovery packets come back with the ports as well as the IP address?

Would much rather use that data and avoid the change to `probe_wemo` if we can.

ausil:
the port seems to be available via basicevent.hostname

so calling 
devices = pywemo.discover_devices()
devices[1].basicevent.hostname
gives '10.46.1.7:49154'

ausil:
with the initial approach to setup home-assistant reported

2017-07-07 17:44:29 INFO (MainThread) [homeassistant.setup] Setup of domain wemo took 38.5 seconds.

with the approach in https://github.com/home-assistant/home-assistant/issues/8414 it reports

2017-07-08 17:27:23 INFO (MainThread) [homeassistant.setup] Setup of domain wemo took 35.6 seconds.

lolento:
Hello,

Has the new probe ports been merged to pywemo?

pavoni:
@lolento - this PR includes other changes as well. 

Fairly happy to merge a PR that just contains the extra ports - what do you need it for?

ausil:
just extending the ports alone is fairly useless, at least for the case of a single ip hosting multiple ports. you have to have something that ties the port and something about it together, or things will just not work. The approach here really is correct for all use cases.

pavoni:
@ausil I'm not keen on the description logic - I think the better approach for your device is to stop pywemo probing / re-probing the device.

pavoni:
Closing since this appears to be stale. Happy to open discussions again.

Issue #80, opened 2017-09-02
Probe additional ports


exxamalte:
I own a WeMo Insight switch, running firmware 'WeMo_WW_2.00.10966.PVT-OWRT-Insight'. It appears in the WeMo app just fine, but Home Assistant cannot find the switch. The switch listens on port `49155` at the moment (i.e. the following URL provides a result: http://<ip>:49155/setup.xml).

The [pywemo library currently probes](https://github.com/pavoni/pywemo/blob/master/pywemo/ouimeaux_device/__init__.py#L17) ports `49153, 49152, 49154`, and it seems as if probing more ports may be required.
Also, see https://github.com/pdumoulin/blinky/blob/master/blinky.py#L20 and https://gist.github.com/pruppert/af7d38cb7b7ca75584ef#file-wemo-py-L19; both probe ports 49151 to 49155.

PR #81, opened 2017-09-02
Add additional commonly used ports for probing

exxamalte:
Fixes pavoni/pywemo#80

Looking at other similar libraries it seems to be consensus that ports 49151, 49155 should also be probed when communicating with WeMo switches.


pavoni:
Thanks - now released as `0.4.20`.

Thanks for the contribution.

PR #83, opened 2017-10-19
Add Wemo Dimmer support

angel12:
I am sure there are things in here that I did that are not right, please let me know!

### DIFF
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -1,0 +1,6 @@
+from .switch import Switch
+
+class Dimmer(Switch):
+
+    def __repr__(self):
+        return '<WeMo Dimmer "{name}">'.format(name=self.name)
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -1,37 +1,13 @@
-from . import Device
+from .switch import Switch
 
-
-class Dimmer(Device):
-
-    def set_state(self, state):
-        """
-        Set the state of this device to on or off.
-        """
-        self.basicevent.SetBinaryState(BinaryState=int(state))
-        self._state = int(state)
-
-    def off(self):
-        """
-        Turn this device off. If already off, will return "Error".
-        """
-        return self.set_state(0)
-
-    def on(self):
-        """
-        Turn this device on. If already on, will return "Error".
-        """
-        return self.set_state(1)
-
-    def toggle(self):
-        """
-        Toggle the dimmer's state.
-        """
-        return self.set_state(not self.get_state())
-		
+class Dimmer(Switch):		
     def set_brightness(self, brightness):
         """
         Set the brightness of this device to an integer between 1-100.
         """
+        if self.get_state() == 0 or None:
+            self.on()
+
         self.basicevent.SetBinaryState(brightness=int(brightness))
         self._brightness = int(brightness)
 
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -28,18 +28,18 @@ class Dimmer(Device):
         """
         return self.set_state(not self.get_state())
 		
-    def set_dimmer(self, dimmer):
+    def set_brightness(self, brightness):
         """
-        Set the dimmer of this device to on or off.
+        Set the brightness of this device to an integer between 1-100.
         """
-        self.basicevent.SetBinaryState(brightness=int(dimmer))
-        self._dimmer = int(dimmer)
+        self.basicevent.SetBinaryState(brightness=int(brightness))
+        self._brightness = int(brightness)
 
     def fifty(self):
         """
         Set the dimmer to 50%.
         """
-        return self.set_dimmer(50)
+        return self.set_brightness(50)
 
     def __repr__(self):
         return '<WeMo Dimmer "{name}">'.format(name=self.name)
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -34,5 +34,8 @@ class Dimmer(Switch):
         self.basicevent.SetBinaryState(brightness=int(brightness))
         self._brightness = int(brightness)
 
+    def subscription_update(self, _type, _param):
+        return False
+
     def __repr__(self):
         return '<WeMo Dimmer "{name}">'.format(name=self.name)
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -1,12 +1,29 @@
 from .switch import Switch
 
-class Dimmer(Switch):		
+class Dimmer(Switch):
+    def __init__(self, *args, **kwargs):
+        Switch.__init__(self, *args, **kwargs)
+        self._brightness = {}
+
+    def get_brightness(self):
+        """
+        Get brightness from device
+        """
+        brightness = self.basicevent.GetBinaryState().get('brightness')
+        self._brightness = brightness
+
     def set_brightness(self, brightness):
         """
         Set the brightness of this device to an integer between 1-100.
+        Setting the brightness does not turn the light on, so we need
+        to check the state of the switch. 
         """
-        if self.get_state() == 0 or None:
-            self.on()
+        if brightness == 0:
+            if self.get_state() != 0:
+                self.off()
+        else:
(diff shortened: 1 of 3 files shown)


pywemo/ouimeaux_device/dimmer.py
```diff
@@ -0,0 +1,39 @@
+from . import Device
+
+
```
pavoni: Please remove extra blank line.

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -0,0 +1,39 @@
...
+class Dimmer(Device):
+
```
pavoni: Not usual to put a blank here - please remove.

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -0,0 +1,39 @@
+from . import Device
+
+
+class Dimmer(Device):
```
pavoni: You can inherit from Switch - and then remove `set_state`, `on`, `off` and `toggle` methods.

To do this import `Switch` rather than `Device`.
`from .switch import Switch`

And then change this line to
`class Dimmer(Switch):`

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -0,0 +1,39 @@
...
+        Toggle the dimmer's state.
+        """
+        return self.set_state(not self.get_state())
+		
+    def set_brightness(self, brightness):
+        """
+        Set the brightness of this device to an integer between 1-100.
+        """
+        self.basicevent.SetBinaryState(brightness=int(brightness))
+        self._brightness = int(brightness)
```
pavoni: It's possible you'll also need to update the on off state - so that the local on/off state works.

Easy to try - set the brightness to '0'  and then call `get_state()` and check it comes back as `0`, then set the brightness to '100'  and then call `get_state()` and check it comes back as `1`.

Assuming that doesn't work - you can just do something like
`0 if brightness == 0 else 1`

pavoni (changes requested):
Looks great.

A couple of minor formatting suggestions, plus if you change the inheritance you can avoid duplicating some of the existing code.

One small local state issue I think you should double check.

Thanks for the contribution, I think you'll be popular with a lot of people.

Let me know if you need help adding to Home Assistant.

(I also change the PR merge target to `master` - hope you don't mind)

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -1,37 +1,13 @@
...
-        Toggle the dimmer's state.
-        """
-        return self.set_state(not self.get_state())
-		
+class Dimmer(Switch):		
     def set_brightness(self, brightness):
         """
         Set the brightness of this device to an integer between 1-100.
         """
+        if self.get_state() == 0 or None:
```
pavoni: Can I check what you want to do here?

Have you found that setting brightness on the device doesn't turn it on?

angel12: setting brightness does not turn it on, this was some trial and error to make sure setting brightness turns device on as well. Sorry, somewhat new to GitHub, trying to get better at comments ;)

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -1,37 +1,13 @@
...
-        return self.set_state(1)
-
-    def toggle(self):
-        """
-        Toggle the dimmer's state.
-        """
-        return self.set_state(not self.get_state())
-		
+class Dimmer(Switch):		
     def set_brightness(self, brightness):
```
pavoni: Might also be worth adding a `get_brightness()` call - I imagine you'll want this for HA. 

You will want to get this from the device so that you have the right value even if it wasn't set via Python.

Also worth checking that if you turn the device on / off via the wemo app and via the self.on() / self.off() - the `get_brightness` call always returns the right number.

Might need to add some explicit setting of `self._brightness`.

pywemo/ouimeaux_device/dimmer.py
pavoni: In that case you may want to do something like:-
```

if brightness == 0:
    if self.get_state() != 0:
        self.off()
else
    if self.get_state() == 0:
        self.on()
```

Assuming you want to do an explicit turn off too.

pywemo/ouimeaux_device/dimmer.py
angel12: How would I pull the value from the device? 

```
>>> device.basicevent.GetBinaryState()
{'fader': '300:-1:1:0:100', 'BinaryState': '1', 'brightness': '75'}
```

pavoni: `device.basicevent.GetBinaryState().get('brightness')`

pavoni: Plus you may want to make it a number 

This is from elsewhere in the code base
````
            try:
                self._state = int(state.get('BinaryState', 0))
            except ValueError:
                self._state = 0
````

Comment (angel12) [reply pywemo/ouimeaux_device/dimmer.py]: I tried and tried to figure your method out, and couldn't get it :( hope this way is acceptable!

File: pywemo/ouimeaux_device/dimmer.py [hunk 7]
```diff
@@ -0,0 +1,32 @@
...
+        Set the brightness of this device to an integer between 1-100.
+        Setting the brightness does not turn the light on, so we need
+        to check the state of the switch. 
+        """
+        if brightness == 0:
+            if self.get_state() != 0:
+                self.off()
+        else:
+            if self.get_state() == 0:
+                    self.on()
```
pavoni: Indented a bit too much.

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -0,0 +1,32 @@
+from .switch import Switch
+
+class Dimmer(Switch):
+    def __init__(self, *args, **kwargs):
+        Switch.__init__(self, *args, **kwargs)
+        self._brightness = {}
```
pavoni: Probably best to set to `None` (see later)

pywemo/ouimeaux_device/dimmer.py
angel12: Good catch! Like I said, new to python, and pretty new to programming

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def get_brightness(self):
+        """
+        Get brightness from device
+        """
+        brightness = self.basicevent.GetBinaryState().get('brightness')
+        self._brightness = brightness
+
```
pavoni: You can make this cache the value like `get_state` - which will stop it always calling the device unless you're explicit. 

Also I think you forgot to return the value.

Lastly - good to make sure the result is an int. I think wemo returns strings.

(I haven't tested this)
```
    def get_brightness(self, force_update):
        """
        Get brightness from device
        """
        if force_update or self._brightness is None:
            try:
                brightness = int(self.basicevent.GetBinaryState().get('brightness'))
            except ValueError:
                brightness = 0

        self._brightness = brightness
        return self._brightness
```

pavoni:
A few small changes - then happy to merge. Thanks again for doing this.

I suspect your HA component will need to poll to get updates from the device.

Happy to help add subscriptions as a future update if you want real time updates to HA.

pywemo/ouimeaux_device/dimmer.py
angel12: should I have to put force_update=False in the first line? Thats the only way I can get that to work

pavoni: You're quite right - sorry.

Last thought is it will make the subscription logic in HA work  better if you add.

```
    def subscription_update(self, _type, _param):
        return False
```

This will tell HA that this object doesn't pull state from the update data.

Sorry for the slight random feeling comments - have been rushing around.

angel12: I just made the changes in my branch, but will test here this afternoon. Thanks for all your help, definitely been a learning experience!

pavoni (approved):

pavoni:
Looks good - let me know when you're happy - and I'll merge and release.

angel12:
Looks like it's good, just tested importing pywemo, as well as get_brightness, set_brightness, on, off, toggle.

pavoni:
Now released as `0.4.21` thanks again for the contribution.

Tag me in your home assistant PR if I can help.

Issue #82, opened 2017-10-02
SSDP discovery fails if string is not valid UTF-8


alanmimms:
Some device in my network has an SSDP response that blows up `ssdp.py` discovery `sock.recv().decode("utf-8")` with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x90 in position 5: invalid start byte`. Since I cannot control all devices and since this is really a robustness issue, I tried replacing the bad UTF-8 with the default replacement by changing the code to
```
                response = sock.recv(1024).decode("UTF-8", "replace")
```
This solves my problem by allowing the discovery to continue even after a response from such an evil device is found. It seems harmless to do this.

alanmimms:
Sorry I've been inattentive. I'll try it today if I can squeeze it in. Thanks for fix.

pavoni:
Thanks @alanmimms - no criticism - I took ages to get to it!

Have merged it as it seemed harmless and I was doing another release.

Terrific if you can check it works OK.

PR #84, opened 2017-10-23
Fix discovery error if string is not valid UTF-8

pavoni:
Closes https://github.com/pavoni/pywemo/issues/82

Fix as suggested in the issue.

@alanmimms sorry for the time taken to do this. Are you able to check it works OK for you?


pavoni:
Released in `0.4.22`

Issue #85, opened 2017-10-24
Issue with Dimmer switch


angel12:
Currently, after calling get_brightness on device once, you cannot call it again. This is the error I am getting now:

```
>>> device.get_brightness()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "*wemo dimmer\pywemo-master\pywemo\ouimeaux_device\dimmer.py", line 18, in get_brightness
    self._brightness = brightness
UnboundLocalError: local variable 'brightness' referenced before assignment
```

pavoni:
You need to put the assignment inside the if test

```
        if force_update or self._brightness is None:
            try:
                brightness = self.basicevent.GetBinaryState().get('brightness')
            except ValueError:
                brightness = 0
            self._brightness = brightness

        return self._brightness
```

Happy to release if you  an test and send a PR.

Sorry I missed this!

angel12:
Hmm, even with those changes I am still getting this error. 

```
>>> device.get_brightness()
'80'
>>> device.set_brightness(25)
>>> device.get_brightness()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "e:\shares\spencer\wemo dimmer\pywemo-master\pywemo\ouimeaux_device\dimmer.py", line 17, in get_brightness
    self._brightness = brightness
UnboundLocalError: local variable 'brightness' referenced before assignment
>>>
```


Current get_brightness:
```
    def get_brightness(self, force_update=False):
        """
        Get Brightness From Device
        """
        if force_update or self._brightness is None:
            try:
                brightness = self.basicevent.GetBinaryState().get('brightness')
            except ValueError:
                brightness = 0
        self._brightness = brightness

        return self._brightness

```


Edit: 
Here it works, kind of:
```
    def get_brightness(self, force_update=False):
        """
        Get Brightness From Device
        """
        global brightness
        if force_update or self._brightness is None:
            try:
                brightness = self.basicevent.GetBinaryState().get('brightness')
            except ValueError:
                level = 0
        self._brightness = brightness
```

```
>>> device.get_brightness()
'25'
>>> device.set_brightness(50)
>>> device.get_brightness()
'25'
>>> device.basicevent.GetBinaryState().get('brightness')
'50'
>>>
```

But the get_brightness function didn't pull the new value, which I am assuming has to do with subscription? I had to get the brightness from scratch, since self._brightness already had the value cached

pavoni:
[...]

PR #86, opened 2017-10-24
Fix for #85

angel12:
Fix for issue #85 

Next stop, subscriptions!

### DIFF
--- a/pywemo/__init__.py
+++ b/pywemo/__init__.py
@@ -5,6 +5,7 @@ Lightweight Python module to discover and control WeMo devices.
 from .ouimeaux_device import Device as WeMoDevice
 from .ouimeaux_device.insight import Insight
 from .ouimeaux_device.lightswitch import LightSwitch
+from .ouimeaux_device.dimmer import Dimmer
 from .ouimeaux_device.motion import Motion
 from .ouimeaux_device.switch import Switch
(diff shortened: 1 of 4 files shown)


pavoni:
This pr doesn't seem to include any changes (except an extra return at the end of the file).

angel12:
derp, thats what I get for not knowing GitHub well enough :( that should have the actual fix on it now

pavoni:
Released in `0.4.22`

Subscriptions might already work in HA, assuming you've picked up the existing HA code for switches etc.

I think wemo subscribes to the event  you want - and returning `False` from `subscription_update ` should tell the HA wrapper code that it needs to refresh the state when it gets an update.

Easy to check - just load in HA - change the brightness in the app and see if HA updates in a second or so.

If it doesn't already work - should be a simple fix.

Next step would be to avoid the extra callback to get device state. Examples of how to do this  are here:
https://github.com/pavoni/pywemo/blob/master/pywemo/ouimeaux_device/insight.py#L20
https://github.com/pavoni/pywemo/blob/master/pywemo/ouimeaux_device/coffeemaker.py#L75

First step is just to trace the parameters - and work out where you can get the data you want. You can see a suitable debug trace in the insight example.

The Switch should already ready be able to handle on/off (if you call it - again see the insight example) - so you'll just need to handle brightness.

Issue #91, opened 2017-12-22
still an issue with humidifier?


johngo7470:
>>> import pywemo
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pywemo/__init__.py", line 14, in <module>
    from .ouimeaux_device.humidifier import Humidifier
  File "pywemo/ouimeaux_device/humidifier.py", line 44, in <module>
    ('100', 4), # "Always On" mode
TypeError: object() takes no parameters
>>>

pavoni:
I think this is a python 2 vs python3 error.

Most users are running this on Python 3 - so we've probably missed a Python 2 problem.

sqldiablo:
I'll see if I can reproduce this in python 2.x and get it fixed. I did all of my testing against 3.6.

johngo7470:
I spun up python3 (3.5.3) and it complains it can't find the netifaces module - guessing netifaces requires 3.6.  I just tried it in 3.6, and when I tried to import pywemo, it's complaining there's no module 'requests'...

johngo7470:
I did an npm install requests, and it indicated that requests@0.2.2 was installed, but python 3.6 can't seem to find it.

pavoni:
Sounds like JavaScript. Try `pip/pip3 install`?

Or to be exact `pip install -r requirements.txt` to install the libraries required.

Thanks to @sqldiablo for sorting the python2 load.

sqldiablo:
I was able to reproduce the humidifier issue in Python 2.7 and determined it was an issue with the way I was defining one of the enums. I've resolved it in my fork and will do a PR to bring it over.

johngo7470:
'pip3 install requests' yielded 'requirement already satisfied...

In python(2), I do progress past the humidifier issue while building, but now trip on the netifaces, again.

A pip install netifaces, took care of that, and I'm cooking on gas! (able to import/run pywemo)

Thanks for the speedy resolution!
JG

sqldiablo:
Excellent, I'm glad to hear that the PR fixed the issue for you.

PR #92, opened 2017-12-22
Humidifier.py - Python 2.x enum fix

sqldiablo:
Changed desired humidity enum to fix issues with Python 2.x.

This should fix the humidifier.py error from Issue #91.


pavoni:
Thanks @sqldiablo - released as 0.4.24

PR #95, opened 2018-03-14
Allow specifying custom ports for probing.

mcspr:
I was having trouble controlling WiFi switches with [custom firmware](https://bitbucket.org/xoseperez/fauxmoesp) emulating WeMo device via Home Assistant. Emulated switches can have multiple virtual switches on the same ip address with non-standart ports eg. - 52000, 52001, 52002. Discovery logic in HA uses `discover_devices` first and correctly finds, for example, 3 switches. But `probe_wemo` will fail next step, even if device's setup and control url from discovery process already has the port.

What I am proposing is to remember port on the device object. Keeping `probe_wemo` as generic host scanner allow specifying custom ports list as an argument, add new `probe_device` which can also include already known port in the scan process. Reconnection process will also first try to use last known port.

For HA part, `probe_wemo` can be replaced with `probe_device` for already discovered devices and allow port definition in static addresses. And just fallback to `probe_wemo` in case there is no port.

References:
https://github.com/home-assistant/home-assistant/pull/8414 old pr that I found looking to solve this
https://github.com/pavoni/pywemo/pull/78 proposed adding multiple new ports, this only adds one per device.

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -19,15 +19,24 @@ log = logging.getLogger(__name__)
 # Start with the most commonly used port
 PROBE_PORTS = (49153, 49152, 49154, 49151, 49155)
 
+def probe_device(device):
+    """Probe a device for available port.
 
-def probe_wemo(host):
+    This is an extension for probe_wemo, also probing current port.
+    """
+    ports = PROBE_PORTS
+    if not device.port in ports
+        ports = (device.port,) + ports
+    return probe_wemo(device.host, ports)
+
+def probe_wemo(host, ports=PROBE_PORTS):
     """Probe a host for the current port.
 
     This probes a host for known-to-be-possible ports and
     returns the one currently in use. If no port is discovered
     then it returns None.
     """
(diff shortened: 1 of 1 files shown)


pavoni:
I think this is a much clearer approach than than suggested previously - and is also better for normal wemo devices - because in a case of failure it will restart with the port that worked last time.

There was a small typo in the PR (a missing `:`) that I fixed.

Thanks for the contribution, now released as 0.4.26.

I'd normally offer to update HA too - but in this case I think you'll also need some HA changes (which I suspect you already have).

Sorry it took me so long to get to this.

mcspr:
Thank you for pointing out about real wemo, I did a double check because of that typo (and because it was 2 months :)
`probe_device` should always prepend port and not only when it is non-standard - #98 

As for HASS - I will double check that too later today with live hass and devices. For now it is here - https://github.com/home-assistant/home-assistant/compare/dev...mcspr:wemo-host-port
Now wemo doesn't re-probe discovered devices and allows host:port in static device list.
Plus, bump to 0.4.26 to support non-standard ports on reconnect - because of possible arbitrary ports.

Probably should add documentation too.

PR #98, opened 2018-05-17
Always scan current port first

mcspr:
Fixup #95 - prepend standard ports too.
+ pep8, use before define.


pavoni:
Thanks for adding this - now released as `0.4.27`

PR #99, opened 2018-06-06
Make probe timeout configurable

pjz:
I've got a wemo on a noisy network; it fails at 10s but works if I bump it up to 45.  It would be nice to surface this so I don't have to hack the module itself.


pavoni:
Good idea - now released as `pywemo-0.4.28` - thanks for the contribution.

PR #100, opened 2018-06-16
Fix available status on light bulbs

gstorer:
My Wemo bulbs don't properly report their status after power up or power down using the 'GetEndDevices' request. 'GetEndDevicesWithStatus' appears to report almost exactly the same thing with a couple of extra xml elements and it does correctly notice the power state change. I have kept the old request in the code as a backup in-case older or other devices don't have 'GetEndDevicesWithStatus'.


pavoni:
Thanks for adding the fallback - can't see any issues with this.

Now released as `pywemo-0.4.28` - thanks for the contribution.

pavoni:
I see you have a pending HA PR - so will leave you to release this to HA - but let me know if you need me to do anything.

PR #104, opened 2018-11-04
Allow autodiscovery of mac in pywemo.discovery.device_from_description()

sqldiablo:
If None is passed to pywemo.discovery.device_from_discovery() for the mac parameter, pywemo will attempt to discover the mac from the description_url.

This will help resolve https://github.com/home-assistant/home-assistant/issues/16010 as Home Assistant calls pywemo.discovery.device_from_description(), passing in None as the mac. This is causing pywemo to never populate the mac of the device, and therefore pywemo can't use the mac during rediscovery when a device goes offline/changes IP address. If pywemo doesn't know the device mac, then rediscovery only attempts to probe the existing device, and doesn't ever run a full discovery to look for the device again (because it needs to find the device by mac to know it has found the same device again).

### DIFF
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -1,6 +1,7 @@
 """
 Module to discover WeMo devices.
 """
+import logging
 import requests
 
 from . import ssdp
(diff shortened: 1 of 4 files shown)


sqldiablo:
Hey @pavoni, can you give this a once-over when you have a chance?

pavoni:
Looks fine to me.

pavoni:
Released as 0.4.29. Thanks for the contribution.

PR #105, opened 2018-11-06
Update rediscovery logic to match on serial instead of only on mac

sqldiablo:
This change allows the rediscovery logic to find a device by serial number as well as mac.


pavoni:
Looks good to me. Thanks for the contribution.

pavoni:
Now released as `0.4.30`

PR #106, opened 2018-11-08
Fix rediscovery logic to match on mac OR serial

sqldiablo:
I accidentally made rediscovery have to match on both mac AND serial. This change makes it so it will rediscover the device if it matches one OR the other. Sorry about that.


pavoni:
Should have spotted that! Will re-release...

sqldiablo:
You know what, this still isn't right. Give me a couple to get that fixed again. This will now false-positive match when searching for a specific device.

sqldiablo:
Apparently ifs are difficult... ;-)

pavoni:
Sure -let me know when you're happy.

PR #107, opened 2018-11-08
Fix false-positive in rediscovery

sqldiablo:
Fixing a false-positive search condition. The logic will now to matching correctly and still match all devices if no search parameters are passed


sqldiablo:
Ok, I think I finally landed on the way I want this logic to work. Sorry for the confusion on this.

sqldiablo:
Closing this so I can fix my fork.

PR #109, opened 2018-11-10
Humidifier - Add method to reset filter life

sqldiablo:
Adds a method to the humdifier device to allow the user to reset the filter life back to 100% (for when you install a new filter).


sqldiablo:
@pavoni, can you please merge this when you get a chance?

PR #110, opened 2018-11-22
Increase the scanned port range

meatheadmike:
After the firmware bump last night (WeMo_WW_2.00.11215.PVT-OWRT-Dimmer) my wemo dimmers are definitely responding on additional ports.

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -17,7 +17,7 @@ from .api.xsd import device as deviceParser
 LOG = logging.getLogger(__name__)
 
 # Start with the most commonly used port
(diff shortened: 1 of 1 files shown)


pavoni:
Thanks for the PR - very happy to add some new ports.

The ports are probed in order - and the probing ends when the device is found, so we'd put the most commonly used ports first to try and speed up the process. 

It probably deserved a comment in the code!

meatheadmike:
That makes sense. Although who knows if Belkin kept the same port logic with this release :-/ I've kept the port ordering from the previous code and added the new ones that it seems to be responding on now.

pavoni:
Thanks for the contribution - now released as `0.4.33`

PR #111, opened 2018-12-15
fixed a bug with rescanning devices when a port number changes

meatheadmike:
See here for a discussion of the issue:

https://github.com/home-assistant/home-assistant/issues/19153

Here's how I can trigger the bug with the current code:
```
>>> import pywemo
>>> ip = '192.168.1.206'
>>> port = pywemo.ouimeaux_device.probe_wemo(ip)
>>> url = f"http://{ip}:{port}/setup.xml"
>>> device = pywemo.discovery.device_from_description(url,None)
>>> device._reconnect_with_device_by_discovery()
Unable to reconnect with Kitchen lights in 5 tries. Stopping.
```

With this PR it now behaves as I would expect:
```
>>> import pywemo
>>> wemo = pywemo.SubscriptionRegistry()
>>> wemo.start()
>>> ip = '192.168.1.206'
>>> port = pywemo.ouimeaux_device.probe_wemo(ip)
>>> url = f"http://{ip}:{port}/setup.xml"
>>> device = pywemo.discovery.device_from_description(url,None)
# force a bad port to duplicate what happens when the port changes:
>>> device.port = 49152
>>> device._reconnect_with_device_by_discovery()
>>> device.port
49153
```

### DIFF
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -252,14 +252,16 @@ def scan(st=None, timeout=DISCOVER_TIMEOUT, max_entries=None, match_mac=None, ma
 
                 # Search for devices
                 if st is not None or match_mac is not None or match_serial is not None:
-                    if st is not None and st == entry.st:
-                        entries.append(entry)
-
-                    if entry not in entries and match_mac is not None and match_mac == mac:
-                        entries.append(entry)
-
-                    if entry not in entries and match_serial is not None and match_serial == serial:
-                        entries.append(entry)
+                    if entry not in entries:
+                        if match_mac is not None:
+                            if match_mac == mac:
+                                entries.append(entry)
+                        elif match_serial is not None:
+                            if match_serial == serial:
+                                entries.append(entry)
+                        elif st is not None:
+                            if st == entry.st:
+                                entries.append(entry)
                 elif entry not in entries:
                     entries.append(entry)
 


sqldiablo:
Thanks, @meatheadmike. I'm testing this locally and will let you know how it works out.

sqldiablo:
Ok, so I've been working on testing this PR, as well as the issue it stems from (https://github.com/home-assistant/home-assistant/issues/19153), and I've been digging into the root cause all day (and into the morning). This PR won't hurt anything, but it won't fix the root of the issue either. I still have issues if a device changes IP more than once (it'll fail to rediscover the second time around), and this points to a fundamental flaw in the way rediscovery is done (I'm looking at you, lines: https://github.com/pavoni/pywemo/blob/2e9260c871491d8956899fc11e95da70befa8f2a/pywemo/ouimeaux_device/__init__.py#L119 and https://github.com/pavoni/pywemo/blob/2e9260c871491d8956899fc11e95da70befa8f2a/pywemo/ouimeaux_device/__init__.py#L149).

I'm working on a pretty sizable set of changes for pywemo to address this and to fix the way rediscovery works (as well as fix the way that device services know what URL/IP/port to use in the first place). It's going to take me at least a couple more days to get this nailed down and submit a PR with my changes.

I have no objection to merging this PR as it stands, but I'll be replacing the changes in this PR with those in a PR of my own later.

meatheadmike:
I can't wait to see what you come up with. For now I've manually installed pywemo on my Hassio installation. My Wemo's are now a lot more reliable. 

I'm not sure if you saw this:
https://github.com/pavoni/pywemo/issues/112

But it looks like a good opportunity for performance improvement.

pavoni:
Sorry for the slow reply.

Generally I'd rather wait until we have a fix you both think solves the whole issue - unless you think the fix is urgent (or the ultimate solution will be difficult).

Thanks for the contribution and analysis on this.

sqldiablo:
I think the fix is urgent, and the one I'm writing is more involved. I'd merge in Mike's PR and cut a new version of pywemo. I can PR the new pywemo version into HA once that's done.

Thanks,
Adam Belebczuk

(Sent from my mobile device)
________________________________

sqldiablo:
Since this PR has been superseded by https://github.com/pavoni/pywemo/pull/113, we should probably close this one.

meatheadmike:
Sounds good to me.

Issue #112, opened 2018-12-15
SSDP scan uses "ssdp:all" instead of "urn:Belkin:service:basicevent:1"


meatheadmike:
SSDP scan basically tells every single upnp device on the network to respond. I can't figure out why this is needed. It certainly adds a lot of unwanted network traffic from what I'm seeing on my end (router, chromecasts, harmony devices, etc). From what I can determine all wemo devices respond with "urn:Belkin:service:basicevent:1". So we should be able to narrow this down significantly. If you're OK with it I can provide a PR. Let me know what you think...

pavoni:
If this works it sounds like a win with no disadvantages (and in fact may help with the odd problem where we've not understood the reply from a non wemo device and crashed!).

So a PR would be very welcome.

sqldiablo:
I'll bake this into my rediscovery fixes and test it along with those changes. I've been annoyed by this as well, so thanks for suggesting the fix for it. :-)

meatheadmike:
Something else to consider while you're working on the big fix - I did a wireshark packet cap on the SSDP traffic to see what what going on during discovery. It turns out the official app actually sends out 2 separate M-SEARCH requests in fairly quick succession once it opens. This kind of makes sense since it's UDP and thus by definition has no delivery guarantee. In practice pywemo tends to under-discover my devices. This is likely due in part to my saturated network. Implementing 2 M-SEARCH requests like the official app seems to help.

sqldiablo:
It turns out that the older WeMo devices (like the WeMo switch, the older, more oval-shaped smart plug) don't respond to the URN search (I tried), and instead only respond to a "upnp:rootdevice" search. I can take out the "ssdp:all" search and replace it with the URN search, but pywemo is using the "upnp:rootdevice" search to find devices, and isn't even sending the "ssdp:all" search during normal discovery.

The fun part is that HA still discovered the device when it ran a discovery scan using its built-in discovery component and doing a netdisco discovery, even when pywemo's discovery scan didn't find the device. When this happened, the device was still added to Home Assistant successfully and worked correctly. But since that is Home Assistant functionality, we can't rely on that in the base pywemo module, so I have to leave it with the "upnp:rootdevice" search.

sqldiablo:
I also tried the double scan, but it gave me back duplicate results, which I don't like. I'm going to leave the second scan out of things for now.

meatheadmike:
[...]

PR #113, opened 2018-12-20
Test urn fix

meatheadmike:
So here's my stab at implementing https://github.com/pavoni/pywemo/issues/112. 

Using the urn of "urn:Belkin:service:basicevent:1" seems to work nicely in my environment. I currently have several wall switches, wall dimmers, plugin wemos (the old-style original) and insight plugins.

All of them are detected correctly using this code.

Note that detection is not perfect every time. I did implement the 2 successive M-SEARCH requests (as I discussed earlier). This mimics what I was seeing in the packet capture from the official app. If this is a concern I can remove it. However for my use case it seems to do a better job at detection.

I managed to avoid duplicate entries by comparing the mac and serial of the entry instead of doing an `entry in entries` comparison.

You'll note that the MX is set to 1 and 2 respectively. This also matches what I was seeing. From my (admittedly limited) understanding of ssdp, MX tells the remote device "respond within this number of seconds". So in the original implementation we were saying "respond within 5 seconds" but we were also timing out in 5 seconds. I imagine if a device decided to respond at the 5 seconds mark and we timed out at 5 seconds then we'd miss the response from said device. So maybe that also explains why I'm seeing more reliability now.

### DIFF
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -198,16 +198,9 @@ def build_ssdp_request(st, ssdp_mx):
         '', '']).encode('ascii')
 
 
-def entry_in_entries(entry, entries):
-    if entry.description is not None:
-        device = entry.description.get('device', {})
-        mac = device.get('macAddress')
-        serial = device.get('serialNumber')
-    else:
-        mac = None
-        serial = None
+def entry_in_entries(entry, entries, mac, serial):
     # If we don't have a mac or serial, let's just compare objects instead:
-    if mac == None and serial == None:
+    if mac is None and serial is None:
         return entry in entries
     for e in entries:
         if e.description is not None:
@@ -282,7 +275,7 @@ def scan(st=None, timeout=DISCOVER_TIMEOUT, max_entries=None, match_mac=None, ma
 
                 # Search for devices
                 if st is not None or match_mac is not None or match_serial is not None:
-                    if not entry_in_entries(entry, entries):
+                    if not entry_in_entries(entry, entries, mac, serial):
                         if match_mac is not None:
                             if match_mac == mac:
                                 entries.append(entry)
@@ -292,7 +285,7 @@ def scan(st=None, timeout=DISCOVER_TIMEOUT, max_entries=None, match_mac=None, ma
                         elif st is not None:
                             if st == entry.st:
                                 entries.append(entry)
-                elif not entry_in_entries(entry, entries):
+                elif not entry_in_entries(entry, entries, mac, serial):
                     entries.append(entry)
 
                 # Return if we've found the max number of devices
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -8,23 +8,19 @@ import logging
 from datetime import datetime, timedelta
 import threading
 import xml.etree.ElementTree as ElementTree
-
+import time
 import requests
 
 from .util import etree_to_dict, interface_addresses
 
-DISCOVER_TIMEOUT = SSDP_MX = 5
+DISCOVER_TIMEOUT = 5
 
 RESPONSE_REGEX = re.compile(r'\n(.*)\: (.*)\r')
 
 MIN_TIME_BETWEEN_SCANS = timedelta(seconds=59)
 
-# Devices and services
-ST_ALL = "ssdp:all"
-
-# Devices only, some devices will only respond to this query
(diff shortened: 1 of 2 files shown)


pavoni:
This looks good to me, and discovers my Wemo Makers, Insight Switches and regular Switches without problems. I think it's as least as reliable as the old code - and probably better.

Although as you say it isn't 100% (and nor was the old code).

@sqldiablo do you have a few minutes to check on your network too?

Otherwise happy to merge and release, thanks for taking the time to put this together.

sqldiablo:
@pavoni, sure, give me a bit to get that setup and put it through its paces.

pywemo/ssdp.py
```diff
@@ -191,6 +187,43 @@ def __repr__(self):
...
+def entry_in_entries(entry, entries):
+    if entry.description is not None:
+        device = entry.description.get('device', {})
+        mac = device.get('macAddress')
+        serial = device.get('serialNumber')
+    else:
+        mac = None
+        serial = None
+    # If we don't have a mac or serial, let's just compare objects instead:
+    if mac == None and serial == None:
```
sqldiablo: I would change these to is None instead of == None. In this case, either should work, but is None is generally considered the best practice and is more efficient than the == None check.

pywemo/ssdp.py
```diff
@@ -252,15 +282,17 @@ def scan(st=None, timeout=DISCOVER_TIMEOUT, max_entries=None, match_mac=None, ma
...
-                    if st is not None and st == entry.st:
-                        entries.append(entry)
-
-                    if entry not in entries and match_mac is not None and match_mac == mac:
-                        entries.append(entry)
-
-                    if entry not in entries and match_serial is not None and match_serial == serial:
-                        entries.append(entry)
-                elif entry not in entries:
+                    if not entry_in_entries(entry, entries):
```
sqldiablo: Since entry_in_entries also turns the entry object into a device object, why not pass in the device object from line 276 to entry_in_entries and use that instead of having entry_in_entries do repetitive work to what has already been done in line 276?

pywemo/ssdp.py
```diff
@@ -252,15 +282,17 @@ def scan(st=None, timeout=DISCOVER_TIMEOUT, max_entries=None, match_mac=None, ma
...
+                        if match_mac is not None:
+                            if match_mac == mac:
+                                entries.append(entry)
+                        elif match_serial is not None:
+                            if match_serial == serial:
+                                entries.append(entry)
+                        elif st is not None:
+                            if st == entry.st:
+                                entries.append(entry)
+                elif not entry_in_entries(entry, entries):
```
sqldiablo: This should probably just be an else. I don't think we need to call entry_in_entries again here.

sqldiablo (changes requested):

sqldiablo:
This seems to be working fine for me with @meatheadmike's updates. It discovered my humidifier, all of the variants of the smart plugs (old one, mini, insight), my dimmer, and the WeMo Link-powered LED lights on the first try (much more consistently than before). I'm fine with merging this once my code review suggestions have been addressed.

pywemo/ssdp.py
meatheadmike: You're right. Normally I run these sorts of things through a pep-8 check.

pywemo/ssdp.py
meatheadmike: If st, match_mac and match_serial are all `None` then we'd have the potential for duplicate entries. That's why I put the check in place. Line 82 seems to have this exact scenario:

https://github.com/pavoni/pywemo/blob/fa15bfe89f2265cbc11c3d815b364099bdb0adeb/pywemo/ssdp.py#L82

pywemo/ssdp.py
meatheadmike: Good call.

sqldiablo (approved):

pywemo/ssdp.py
sqldiablo: Good call, I didn't think about that.

pavoni:
Now released as `0.4.34`. Thanks for the contribution.

sqldiablo:
Thanks, guys. I'll do a PR to roll this into the next release of HA.

sqldiablo:
Also impacts home-assistant/home-assistant#19153

PR #125, opened 2019-01-03
Allow disabling of rediscovery

sqldiablo:
## Description:
This change allows a user to disable pywemo's rediscovery feature. This will be used by a future change to Home Assistant so that Home Assistant can use its own rediscovery logic.

## BREAKING CHANGE:
If calling a device action (such as to get the state, set the state, etc. fails, it will now raise an ActionException that will be returned to the caller (even outside of pywemo). This was needed so that Home Assistant can catch the error and know that a device action\update has failed.

## Checklist:


pavoni:
Looks like you may need to rebase this (I think it has picked up the commits you already merged)

sqldiablo:
Yeah, this one turned out a bit different than expected, since I thought I had rebased it...lol I'm still getting rebasing figured out.

sqldiablo:
Ok, I got a way cleaner PR for this closing this one.

PR #127, opened 2019-01-04
Allow Disabling Rediscovery

sqldiablo:
## Description:
This change allows a user to disable pywemo's rediscovery feature. This will be used by a future change to Home Assistant so that Home Assistant can use its own rediscovery logic.

## BREAKING CHANGE:
If calling a device action (such as to get the state, set the state, etc. fails, it will now raise an ActionException that will be returned to the caller (even outside of pywemo). This was needed so that Home Assistant can catch the error and know that a device action\update has failed.

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/api/service.py
+++ b/pywemo/ouimeaux_device/api/service.py
@@ -77,8 +77,9 @@ class Action:
                     response_dict[response_item.tag] = response_item.text
                 return response_dict
             except requests.exceptions.RequestException:
-                LOG.warning("Error communicating with %s, retry %i",
-                            self._device.name, attempt)
+                LOG.warning("Error communicating with %s at %s:%i, retry %i",
+                            self._device.name, self._device.host,
+                            self._device.port, attempt)
 
                 if self._device.rediscovery_enabled:
                     self._device.reconnect_with_device()
--- a/pywemo/ouimeaux_device/api/service.py
+++ b/pywemo/ouimeaux_device/api/service.py
@@ -1,4 +1,5 @@
 """Representation of Services and Actions for WeMo devices."""
+# flake8: noqa E501
 import logging
 from xml.etree import cElementTree as et
 
--- a/pywemo/ouimeaux_device/api/service.py
+++ b/pywemo/ouimeaux_device/api/service.py
@@ -1,3 +1,4 @@
+"""Representation of Services and Actions for WeMo devices."""
 import logging
 from xml.etree import cElementTree as et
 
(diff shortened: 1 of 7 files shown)


pywemo/ouimeaux_device/api/service.py
```diff
@@ -6,39 +7,56 @@
 from .xsd import service as serviceParser
 
 
-log = logging.getLogger(__name__)
+LOG = logging.getLogger(__name__)
+MAX_RETRIES = 3
 
+# The formatting here is a very hacky way around a too-long line flake8 error
```
pavoni: I think it's possible to turn off flake checking for this http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html

I think the previous code is clearer (and from your comment I think you agree!).

pywemo/ouimeaux_device/bridge.py
```diff
@@ -52,35 +52,36 @@ def __repr__(self):
...
-        end_device_list = et.fromstring(end_devices.get('DeviceLists'))
-
-        for light in end_device_list.iter('DeviceInfo'):
-            # pylint: disable=invalid-name
-            uniqueID = light.find('DeviceID').text
-            if uniqueID in self.Lights:
-                self.Lights[uniqueID].update_state(light)
-            else:
-                self.Lights[uniqueID] = Light(self, light)
+        if force_update or self.Lights is None or self.Groups is None:
```
pavoni, line 58: Why did you want to add `force_update` as part of this PR?

pavoni (changes requested):
Looks good to me. 

One suggestion, and one question.

BTW now you have access to the repo - you can just branch there - rather than needing a local repo.

Seemed to work OK with my Makers.

pywemo/ouimeaux_device/bridge.py
sqldiablo: I was bringing the bridge in-line with the other devices for this. I can default it to True, to make backwards compatibility a little safer.

pywemo/ouimeaux_device/api/service.py
sqldiablo: Yeah, this was a pain to deal with. You usually use an inline "# noqa" comment to ignore stuff like this, but being a multi-line string, that didn't work. I'll keep checking to see if I missed the way to do it.

sqldiablo:
As for branching in this repo, I'd feel a little safer if I work in my fork and merge back into here, unless you object. I didn't want to muddy things up in this repo too much or cause an issue I can't cleanly fix. At least with my fork, I can always delete and recreate it if things get too out of hand.

pavoni:
Completely your choice - but I always found keeping a fork in step to be a pain.

sqldiablo:
Now that I actually know how (rebase wasn't my friend for a while), I've created a couple scripts that help me keep current, so it's not too bad.

pywemo/ouimeaux_device/api/service.py
sqldiablo: Yeah, this is a known bug in flake8 (https://gitlab.com/pycqa/flake8/issues/375). I can ignore the long-line error in the entire file and then revert this back to how it was.

pavoni (approved):

pywemo/ouimeaux_device/bridge.py
sqldiablo: I went ahead and defaulted it to True so it will behave the same if the parameter is omitted as it did before this change.

sqldiablo:
Linking this to #125, which this PR was a replacement for.

PR #142, opened 2019-11-27
Device type is a switch not a humidifier

bpaauwe:
Set the device type string to a more correct value for this type of device.

## Description:

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/switch.py
+++ b/pywemo/ouimeaux_device/switch.py
@@ -31,4 +31,4 @@ class Switch(Device):
     @property
     def device_type(self):
         """Return what kind of WeMo this device is."""
-        return "Humidifier"
+        return "Switch"


sqldiablo:
I'm not sure I agree with this change. All of the devices have descriptive device types like CoffeeMaker, Humidifier, Motion, and so on. @pavoni, what do you think?

@bpaauwe, what is the reason you would like to make this change? Is this helping to fix an issue with something that is using this module?

bpaauwe:
@sqldiablo did you actually look at the change?  It looks like commit 952d6edff48011db8ae21df2ae2b591fd14063f3 - Add device_type property to device classes was done by pasting the same code in to each device file and then changing the device type appropriately, but the switch device file didn't get updated after the paste.  

Yes, without fixing this it does cause issues because you can't identify switches by device_type, they show up as humidifiers.  I'm trying to update https://github.com/rl1131/udi-wemo-poly to support more device types, specifically the Wemo Mini switch.

sqldiablo:
Bob, I apologize. I saw the Humidifier value and thought it was for the Humidifier device, and not the switch device. My mistake. This is approved, and I will merge it in for you. Thanks for catching this.

sqldiablo:
Published as version 0.4.41

bpaauwe:
Thanks @sqldiablo!

pavoni:
Thanks to both of you for contributing and merging. Can't believe we missed it for so long!

PR #153, opened 2020-05-27
Add tests for ouimeaux_device.service.api.Action class

sullivanmj:
## Description:
With the eventual changes to pyWeMo of deprecating python 2 support and adding async functionality, I started adding some unit tests, as well as integrating them with the Travis CI. I had some thoughts & questions along the way. I figured it would be easier to discuss these things if there was a working pull request rather than just opening another issue. Especially since adding the first tests will, set precedent for tests that follow. This is the first of hopefully many PRs to add more tests so that we can make sure we're not breaking things along the way.

Anyway, in no particular order some possible points of discussion:

- It appears there was some previous configuration for running pytest with Travis CI, so I was able to reuse that without making too significant of changes to the Travis CI & Tox configuration.
- It seems there are quite a few ways to organize test packages/modules with respect to the code that they test. It was mostly a coinflip for me as to which direction to go, so I'm open to changing that approach.
- I took the liberty of excluding the tests from the existing flake8 call, and adding a separate call to that specific directory with some exclusions for docstrings. I understand if you don't want that somewhat unnecessary complexity in the build.
- For some reason, on my machine, flake8 seemed to be picking up its configuration from tox.ini, where .venv was not excluded - making it take quite a while to execute. This did not appear to be happening on the Travis CI. Perhaps it pulls flake8 config from setup.cfg instead? Anyway, that's why I added that exclusion from the tox.ini - to prevent the pain locally.
- I attempted to squash my commits into some more intelligible/less noisy commits, but now that I see them in the PR form, it might not have worked? Maybe that's just how github works, or maybe I did it wrong, or maybe something else? Either way - I'm not sure if it's a big deal or not.

**Related issue (if applicable):** n/a

## Checklist:

### DIFF
--- a/tests/ouimeaux_device/api/unit/test_service.py
+++ b/tests/ouimeaux_device/api/unit/test_service.py
@@ -1,0 +1,144 @@
+"""Tests for pywemo.ouimeaux_device.api.service."""
+
+from xml.etree import cElementTree as cet
+from xml.etree import ElementTree
+import unittest.mock as mock
+import pytest
+import requests
+
+import pywemo.ouimeaux_device.api.service as svc
+
+HEADERS_KWARG_KEY = 'headers'
+CONTENT_TYPE_KEY = 'Content-Type'
+SOAPACTION_KEY = 'SOAPACTION'
+
+svc.LOG = mock.Mock()
+
+
+class TestAction:
+    @staticmethod
+    def get_mock_action(name="", service_type="", url=""):
+        device = mock.Mock()
+
+        service = mock.Mock()
+        service.serviceType = service_type
+        service.controlURL = url
+
+        action_config = mock.MagicMock()
+        action_config.get_name = lambda: name
+
+        return svc.Action(device, service, action_config)
+
+    def test_call_post_request_is_made_exactly_once_when_successful(self):
+        action = self.get_mock_action()
+        requests.post = mock.Mock()
+        cet.fromstring = mock.MagicMock()
+
+        action()
+
+        requests.post.assert_called_once()
+
+    def test_call_request_has_well_formed_xml_body(self):
+        action = self.get_mock_action(name="cool_name", service_type="service")
+        requests.post = post_mock = mock.Mock()
+        cet.fromstring = mock.MagicMock()
+
+        action()
+
+        body = post_mock.call_args.args[1]
+        ElementTree.fromstring(body)  # will throw if xml is malformed
+
+    def test_call_request_has_correct_header_keys(self):
+        action = self.get_mock_action()
+        requests.post = post_mock = mock.Mock()
+        cet.fromstring = mock.MagicMock()
+
+        action()
+
+        headers = post_mock.call_args.kwargs[HEADERS_KWARG_KEY]
+        for header in [CONTENT_TYPE_KEY, SOAPACTION_KEY]:
+            assert header in headers
+
+    def test_call_headers_has_correct_content_type(self):
+        action = self.get_mock_action()
+        requests.post = post_mock = mock.Mock()
+        cet.fromstring = mock.MagicMock()
+
+        action()
+
+        headers = post_mock.call_args.kwargs[HEADERS_KWARG_KEY]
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -2,7 +2,7 @@
 envlist = lint, pylint
 skip_missing_interpreters = True
 
-[testenv]
+[testenv:pytest]
```
sqldiablo: Can we keep this named [testenv]? That will line up with some other changes I was planning to make to our testing/linting setup.

sqldiablo (changes requested):

tests/ouimeaux_device/api/unit/test_service.py
```diff
@@ -0,0 +1,146 @@
...
+
+from xml.etree import cElementTree as cet
+from xml.etree import ElementTree
+import unittest.mock as mock
+import pytest
+import requests
+
+import pywemo.ouimeaux_device.api.service as svc
+
+HEADERS_KWARG_KEY = 'headers'
```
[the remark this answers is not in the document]
sqldiablo: Can you run black against this?

black --fast pywemo tests

tests/ouimeaux_device/api/unit/test_service.py
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            action()
+        except svc.ActionException:
+            pass
+
+        assert post_mock.call_count == svc.MAX_RETRIES
+
+    def test_call_throws_when_final_retry_fails(self):
+        action = self.get_mock_action()
+        requests.post = mock.Mock(side_effect=requests.exceptions.RequestException)
```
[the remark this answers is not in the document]
sullivanmj: Ha! Black made this line too long. Go figure.

tests/ouimeaux_device/api/unit/test_service.py
```diff
@@ -0,0 +1,148 @@
...
+        try:
+            action()
+        except svc.ActionException:
+            pass
+
+        assert post_mock.call_count == svc.MAX_RETRIES
+
+    def test_call_throws_when_final_retry_fails(self):
+        action = self.get_mock_action()
+        requests.post = mock.Mock(
```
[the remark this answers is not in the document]
sullivanmj: Apparently black/hound are at odds on line length... See [this comment](https://github.com/pavoni/pywemo/pull/153#pullrequestreview-421528798).

tox.ini
sullivanmj: Sure. That was the only way that I could figure out how to get pytest to run in the Travis CI build with the existing Tox configuration. Could you suggest how to get that working with the original configuration here? I haven't worked with Tox before now.

tests/ouimeaux_device/api/unit/test_service.py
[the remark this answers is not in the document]
sqldiablo: lol, that's a good time. Go ahead and edit .hound.yml and set enabled to false for now. I'm more concerned with flake8 and the other tests that Travis runs for us.

tox.ini
sqldiablo: Try adding py35 to the envlist.

sqldiablo: Let's try making it look like this:

```ini
[testenv]
basepython = {env:PYTHON3_PATH:python3}
commands =
     pytest {posargs}
     {toxinidir}/script/check_dirty
deps =
     -r{toxinidir}/requirements_test.txt
     pytest-xdist
```

sqldiablo: Scratch that, you found the issue.

sqldiablo:
This looks good to me, if it looks good to you at this point.

sqldiablo (approved):

sullivanmj:
Yeah, as long as you're OK with the direction I went with setting up the tests, I'm happy with it.

sqldiablo:
I don't have much experience with tests in python because I'm mainly just a hobbyist python developer. What you did seems to make sense, so I'm ok with it. I appreciate your help with getting us started with tests.

PR #157, opened 2020-07-08
Fix parse_basic_state for Wemo LightSwitch.

esev:
## Description:

Subscriptions from Wemo switches return notify results in this format:

```
<e:propertyset xmlns:e="urn:schemas-upnp-org:event-1-0">
<e:property>
<BinaryState>1</BinaryState>
</e:property>
</e:propertyset>
```

This is incompatible with the Device.parse_basic_state method which is
expecting to receive 10 pipe separated parameters.

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -179,22 +179,12 @@ class Device(object):
 
     def parse_basic_state(self, params):
         """Parse the basic state response from the device."""
-        # BinaryState
-        # 1|1492338954|0|922|14195|1209600|0|940670|15213709|227088884
-        (
-            state,  # 0 if off, 1 if on,
-            _x1,
-            _x2,
-            _x3,
-            _x4,
-            _x5,
-            _x6,
-            _x7,
-            _x8,
-            _x9
-        ) = params.split('|')
-
-        return {'state': state}
+        # The BinaryState `params` could have two different formats:
+        #   1|1492338954|0|922|14195|1209600|0|940670|15213709|227088884
+        #   1
+        # In both formats, the first integer value indicates the state.
+        # 0 if off, 1 if on,
+        return {'state': params.split('|')[0]}
 
     def update_binary_state(self):
         """Update the cached copy of the basic state response."""
@@ -208,6 +198,8 @@ class Device(object):
             try:
                 self._state = int(self.parse_basic_state(_params).get("state"))
             except ValueError:
+                LOG.warning("Unexpected BinaryState value `%s` for device %s. "
+                            "Defaulting to _state=0.", _params, self.name)
                 self._state = 0
             return True
         return False
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -198,9 +198,8 @@ class Device(object):
             try:
                 self._state = int(self.parse_basic_state(_params).get("state"))
             except ValueError:
-                LOG.warning("Unexpected BinaryState value `%s` for device %s. "
-                            "Defaulting to _state=0.", _params, self.name)
-                self._state = 0
+                LOG.error("Unexpected BinaryState value `%s` for device %s.",
+                          _params, self.name)
             return True
         return False
 
--- a/pywemo/ouimeaux_device/lightswitch.py
+++ b/pywemo/ouimeaux_device/lightswitch.py
@@ -1,9 +1,26 @@
-"""Representation of a WeMo Motion device."""
+"""Representation of a WeMo LightSwitch device."""
+
+import logging
 from .switch import Switch
+
+LOG = logging.getLogger(__name__)
 
 
 class LightSwitch(Switch):
-    """Representation of a WeMo Motion device."""
+    """Representation of a WeMo LightSwitch device."""
+
+    def subscription_update(self, _type, _params):
+        """Update device state based on subscription event."""
+        LOG.debug("subscription_update %s %s", _type, _params)
+        if _type == "BinaryState":
+            try:
+                self._state = int(_params)
+            except ValueError:
+                LOG.error(
+                    "Could not parse BinaryState update '%s': %s", _params, e)
+                self._state = 0
+            return True
+        return False
 
     def __repr__(self):
         """Return a string representation of the device."""
--- a/pywemo/ouimeaux_device/lightswitch.py
+++ b/pywemo/ouimeaux_device/lightswitch.py
@@ -1,26 +1,9 @@
-"""Representation of a WeMo LightSwitch device."""
-
-import logging
+"""Representation of a WeMo Motion device."""
 from .switch import Switch
-
-LOG = logging.getLogger(__name__)
 
 
 class LightSwitch(Switch):
-    """Representation of a WeMo LightSwitch device."""
-
-    def subscription_update(self, _type, _params):
-        """Update device state based on subscription event."""
-        LOG.debug("subscription_update %s %s", _type, _params)
-        if _type == "BinaryState":
-            try:
-                self._state = int(_params)
-            except ValueError as e:
-                LOG.error(
-                    "Could not parse BinaryState update '%s': %s", _params, e)
-                self._state = 0
-            return True
-        return False
+    """Representation of a WeMo Motion device."""
 
     def __repr__(self):
         """Return a string representation of the device."""
--- a/pywemo/ouimeaux_device/lightswitch.py
+++ b/pywemo/ouimeaux_device/lightswitch.py
@@ -15,7 +15,7 @@ class LightSwitch(Switch):
         if _type == "BinaryState":
             try:
                 self._state = int(_params)
-            except ValueError:
+            except ValueError as e:
                 LOG.error(
                     "Could not parse BinaryState update '%s': %s", _params, e)
                 self._state = 0
--- a/pywemo/ouimeaux_device/switch.py
+++ b/pywemo/ouimeaux_device/switch.py
@@ -1,9 +1,5 @@
 """Representation of a WeMo Switch device."""
-
-import logging
 from . import Device
-
-LOG = logging.getLogger(__name__)
 
 
 class Switch(Device):
@@ -27,15 +23,6 @@ class Switch(Device):
     def toggle(self):
         """Toggle the switch's state."""
         return self.set_state(not self.get_state())
-
-    def parse_basic_state(self, params):
-        """Parse the basic state response from the device."""
-        try:
-            state = int(params)
-        except ValueError as e:
-            state = 0
-            LOG.error("Could not parse state: %s", e)
-        return {"state": state}
 
     def __repr__(self):
         """Return a string representation of the device."""
--- a/pywemo/ouimeaux_device/switch.py
+++ b/pywemo/ouimeaux_device/switch.py
@@ -1,5 +1,9 @@
 """Representation of a WeMo Switch device."""
+
+import logging
 from . import Device
+
+LOG = logging.getLogger(__name__)
 
 
 class Switch(Device):
@@ -23,6 +27,15 @@ class Switch(Device):
     def toggle(self):
         """Toggle the switch's state."""
         return self.set_state(not self.get_state())
+
+    def parse_basic_state(self, params):
+        """Parse the basic state response from the device."""
+        try:
+            state = int(params)
+        except ValueError as e:
+            state = 0
+            LOG.error("Could not parse state: %s", e)
+        return {"state": state}
 
     def __repr__(self):
         """Return a string representation of the device."""


pavoni:
My guess is that your Wemo switch device is not the same as mine, as the current code works OK!

What sort of Wemo device are you using? It may be that we need to distinguish it - and then add a new class to handle it - including the code here.

esev:
Great point. I remember the Wemo app gave a choice of two different switch types.

This is the switch that I have: https://www.belkin.com/us/p/P-WLS040/

The /setup.xml from mine is below. Is the `<hwVersion>v2</hwVersion>` different from yours?

```
<?xml version="1.0"?>
<root xmlns="urn:Belkin:device-1-0">
  <specVersion>
    <major>1</major>
    <minor>0</minor>
  </specVersion>
  <device>
<deviceType>urn:Belkin:device:lightswitch:1</deviceType>
<friendlyName>Porch Light</friendlyName>
    <manufacturer>Belkin International Inc.</manufacturer>
    <manufacturerURL>http://www.belkin.com</manufacturerURL>
    <modelDescription>Belkin Plugin Socket 1.0</modelDescription>
    <modelName>LightSwitch</modelName>
    <modelNumber>1.0</modelNumber>
<hwVersion>v2</hwVersion>
    <modelURL>http://www.belkin.com/plugin/</modelURL>
<serialNumber>2279A029N0XXXX</serialNumber>
<UDN>uuid:Lightswitch-2_0-2279A029N0XXXX</UDN>
    <UPC>123456789</UPC>
<macAddress>3023030XXXX</macAddress>
<hkSetupCode>331-9X-XXX</hkSetupCode>
<firmwareVersion>WeMo_WW_2.00.11451.PVT-OWRT-LIGHTV2</firmwareVersion>
<iconVersion>8|49153</iconVersion>
<binaryState>0</binaryState>
    <new_algo>1</new_algo>
    <iconList>
      <icon>
        <mimetype>jpg</mimetype>
        <width>100</width>
        <height>100</height>
        <depth>100</depth>
         <url>icon.jpg</url>
      </icon>
    </iconList>
    <serviceList>
      <service>
        <serviceType>urn:Belkin:service:WiFiSetup:1</serviceType>
        <serviceId>urn:Belkin:serviceId:WiFiSetup1</serviceId>
        <controlURL>/upnp/control/WiFiSetup1</controlURL>
        <eventSubURL>/upnp/event/WiFiSetup1</eventSubURL>
        <SCPDURL>/setupservice.xml</SCPDURL>
      </service>
      <service>
        <serviceType>urn:Belkin:service:timesync:1</serviceType>
        <serviceId>urn:Belkin:serviceId:timesync1</serviceId>
        <controlURL>/upnp/control/timesync1</controlURL>
        <eventSubURL>/upnp/event/timesync1</eventSubURL>
        <SCPDURL>/timesyncservice.xml</SCPDURL>
      </service>
      <service>
        <serviceType>urn:Belkin:service:basicevent:1</serviceType>
        <serviceId>urn:Belkin:serviceId:basicevent1</serviceId>
        <controlURL>/upnp/control/basicevent1</controlURL>
        <eventSubURL>/upnp/event/basicevent1</eventSubURL>
        <SCPDURL>/eventservice.xml</SCPDURL>
      </service>
      <service>
        <serviceType>urn:Belkin:service:firmwareupdate:1</serviceType>
        <serviceId>urn:Belkin:serviceId:firmwareupdate1</serviceId>
        <controlURL>/upnp/control/firmwareupdate1</controlURL>
        <eventSubURL>/upnp/event/firmwareupdate1</eventSubURL>
        <SCPDURL>/firmwareupdate.xml</SCPDURL>
      </service>
      <service>
        <serviceType>urn:Belkin:service:rules:1</serviceType>
        <serviceId>urn:Belkin:serviceId:rules1</serviceId>
        <controlURL>/upnp/control/rules1</controlURL>
        <eventSubURL>/upnp/event/rules1</eventSubURL>
        <SCPDURL>/rulesservice.xml</SCPDURL>
      </service>

      <service>
        <serviceType>urn:Belkin:service:metainfo:1</serviceType>
        <serviceId>urn:Belkin:serviceId:metainfo1</serviceId>
        <controlURL>/upnp/control/metainfo1</controlURL>
        <eventSubURL>/upnp/event/metainfo1</eventSubURL>
        <SCPDURL>/metainfoservice.xml</SCPDURL>
      </service>

      <service>
        <serviceType>urn:Belkin:service:remoteaccess:1</serviceType>
        <serviceId>urn:Belkin:serviceId:remoteaccess1</serviceId>
        <controlURL>/upnp/control/remoteaccess1</controlURL>
        <eventSubURL>/upnp/event/remoteaccess1</eventSubURL>
        <SCPDURL>/remoteaccess.xml</SCPDURL>
      </service>

      <service>
        <serviceType>urn:Belkin:service:deviceinfo:1</serviceType>
        <serviceId>urn:Belkin:serviceId:deviceinfo1</serviceId>
        <controlURL>/upnp/control/deviceinfo1</controlURL>
        <eventSubURL>/upnp/event/deviceinfo1</eventSubURL>
        <SCPDURL>/deviceinfoservice.xml</SCPDURL>
      </service>
           
      <service>
        <serviceType>urn:Belkin:service:smartsetup:1</serviceType>
        <serviceId>urn:Belkin:serviceId:smartsetup1</serviceId>
        <controlURL>/upnp/control/smartsetup1</controlURL>
        <eventSubURL>/upnp/event/smartsetup1</eventSubURL>
        <SCPDURL>/smartsetup.xml</SCPDURL>
      </service>

      <service>
        <serviceType>urn:Belkin:service:manufacture:1</serviceType>
        <serviceId>urn:Belkin:serviceId:manufacture1</serviceId>
        <controlURL>/upnp/control/manufacture1</controlURL>
        <eventSubURL>/upnp/event/manufacture1</eventSubURL>
        <SCPDURL>/manufacture.xml</SCPDURL>
      </service>

    </serviceList>
   <presentationURL>/pluginpres.html</presentationURL>
</device>
</root>
```

pavoni:
Yes - mine is like this.

https://www.belkin.com/uk/p/P-F7C027/

Looking at the the XML I think what you have is a *LightSwitch* rather than a *Switch*.

Is pywemo discovering it as a `Switch` class?

Wouldn't surprise me if the subscription code was wrong for the *LightSwitch* - I've never seen one - don't think they were every sold in the UK, 

Although I am slightly surprised that a `BasicState` update is returning XML. 

Would be worth setting the log level and getting the trace from this line https://github.com/pavoni/pywemo/blob/master/pywemo/ouimeaux_device/__init__.py#L206

esev:
> Looking at the the XML I think what you have is a _LightSwitch_ rather than a _Switch_.
> 
> Is pywemo discovering it as a `Switch` class?
> 
> Wouldn't surprise me if the subscription code was wrong for the _LightSwitch_ - I've never seen one - don't think they were every sold in the UK,

Ooof. You're totally right here. I hadn't noticed that it was getting detected as a LightSwitch. Thanks for the pointer.

> 
> Although I am slightly surprised that a `BasicState` update is returning XML.
> 
> Would be worth setting the log level and getting the trace from this line https://github.com/pavoni/pywemo/blob/master/pywemo/ouimeaux_device/__init__.py#L206

I had grabbed the NOTIFY XML from tcpdump. Adding the log line you suggested gives this:

```
INFO:pywemo.ouimeaux_device:subscription_update BinaryState 0
```

```
INFO:pywemo.ouimeaux_device:subscription_update BinaryState 1
```

How about I move the code I added to switch.py over to lightswitch.py? Does that sound good to you?

pavoni:
Certainly a good idea to move it!

I think it would probably be cleaner (and more obvious for the next developer) if you override `subscription_update` rather than `parse_basic_state`. @sqldiablo do you have a view?

It would be good to find a way to test this, the subscription code is tricky because of the environment, but you should be able to use a local version of the `pywemo` library if you're using Home Assistant (which I'd guess you are). You can then turn up the log level for `pywemo` and add some debug logs to check the updates are arriving and being processed as you expect.

esev:
I've been using this little script to test. But I'll patch into Home Assistant too and test there too.

```python
import logging
import pywemo
import time

def _cb(_device, _type, _params):
  logging.info('%s %s %s', _type, _params, _device.subscription_update(_type, _params))

sub = pywemo.SubscriptionRegistry()
sub.start()

address = '' # IP address goes here
port = pywemo.ouimeaux_device.probe_wemo(address)
url = 'http://%s:%i/setup.xml' % (address, port)
device = pywemo.discovery.device_from_description(url, None)
print(device)

sub.on(device, None, _cb)
sub.register(device)
try:
  sub.join()
except KeyboardInterrupt:
  sub.stop()
  sub.join()
```

pavoni:
Cool script (we should add that as a demo).

So now I'm puzzled how the code worked in the wrong class?

esev:
> So now I'm puzzled how the code worked in the wrong class?

I was too at first. Switch is the base class for LightSwitch.

pavoni:
Got it  - thanks. Should have realised!

esev:
I've confirmed that this works in Home Assistant too.

I have a Node-RED automation that looks for the light switch to turn on and back off quickly (within 1.5 seconds). This then turns off the rest of the lights in the room. That wasn't working prior to this change and is working now with this change.

Quick question: The `Device.subscription_update` method sets `self._state = 0` when a ValueError is rasied. I copied this logic over to LightSwitch. But I'm wondering if that should just `return False` instead. My understanding is that returning False will signal that there was a state change, but the new state was not understood. Is that an accurate interpretation? What do you think?

pavoni:
The idea was that  the `subscription_update` method should return whether it dealt with the update and so if it didn't a controlling process could force a poll of the device. 

Where the event type should have been understood - but there was an error it's tough to say what the best return would be. My feeling is that it should probably return `True` since the event was understood.

It was part of a change where I hoped to get the code to subscribe to multiple event types - where this mattered more. Handling the subscriptions re-request logic was larger change that I had time for - so I never got this to work.

esev:
> My feeling is that it should probably return True since the event was understood.

That sounds good to me.

Do you know anyone with the older model LightSwitch that could test this change? I think that'd be this one: https://www.belkin.com/us/p/P-F7C030/

I could also use `self._state = int(_params.split('|')[0])` if the older version returned pipe separated values.

sqldiablo:
I've actually got one of these (model F7C030fc) new in box, but never took the time to add it to my debug rig. I should be able to do that this weekend to give this a proper testing. Sorry I'm just now getting to this, it was a busy day at work today.

esev:
Thank you for offering to test. I'd rather have the code correct the first time then break existing devices. No rush.

esev:
I'm interested to get your thoughts on an alternative to this PR:
https://github.com/esev/pywemo/commit/3df7bfe31b94d754bc7f39744059420b3db4bf5d

All of the _x[1-9] variables are unused, so that commit should be backward compatible with existing devices. That commit also works for both LightSwitch devices and the new WiFi Smart Plug devices. https://www.belkin.com/us/p/P-WSP080/ (which is detected as a Switch class and is not fixed with this PR as-is).

esev:
I linked a few more issues from Home Assistant to this PR. It seems like more than just LightSwitch class devices return a BinaryState value of a single integer (0 or 1) rather than 10 pipe separated values. Some Switch (https://www.belkin.com/us/p/P-WSP080/) & Motion devices do also. As this PR won't address the issue with those devices, I think this PR may not be the correct fix. Please let me know what you think of https://github.com/esev/pywemo/commit/3df7bfe31b94d754bc7f39744059420b3db4bf5d, as I think it is the simpler solution and may address the issues found in other devices also.

pavoni:
@sqldiablo What do you think?

esev:
I've tested this new update with the following devices:

1. [WLS040 Smart Light Switch](https://www.belkin.com/us/p/P-WLS040/)
2. [WLS0403 Smart Light Switch 3-Way](https://www.belkin.com/us/p/P-WLS0403/)
3. [F7C059 Smart Dimmer](https://www.belkin.com/us/p/P-F7C059/)
4. [WSP080 Smart Plug](https://www.belkin.com/us/p/P-WSP080/)
5. [F7C029 Insight Smart Plug](https://www.belkin.com/us/p/P-F7C029/)

There is some evidence in https://github.com/home-assistant/core/issues/37544 that the fix works for the Wemo motion sensor (testing ongoing). The debug logs on that issue show that device is also providing a single integer for the update.

I haven't tested these:

1. [F7C027 Switch](https://www.belkin.com/uk/p/P-F7C027/)
2. [F7C063 Mini Smart Plug](https://www.belkin.com/us/p/P-F7C063/)
3. [F7C030 Smart Light Switch](https://www.belkin.com/us/p/P-F7C030/)

esev:
@sqldiablo what do you think about removing the ValueError exception handler in `update_binary_state`? That, or rasing a new UnexpectedBinaryState Exception? It would be unexpected that the BinaryState could not be parsed. My thought is that the exception handler made it hard to discover the issue with the notifications from devices that only send a single integer. There is a warning log there now, but it still kind of hides the unexpected situation.

My concern would be with testing. I don't have access to the Humidifier, Coffee Maker, or a few of the switch / light switch devices so I can't tell if this PR handles all the cases for these devices. Maybe it's okay to break them though - so we are aware of issues sooner.

pavoni:
Will leave @sqldiablo to comment too - but I'd be cautious about removing the exception (although certainly agree that it should have notified better - perhaps as an error).

The issue is that in HA an exception on the monitoring thread will terminate realtime monitoring for all wemo devices until HA is restarted. Quite a few of the devices don't poll - so it would cause the devices to go stale. For the many people who don't look at their log this would cause more problems!

esev:
@pavoni Thank you for providing the bigger picture. I'm pretty new to HA and pywemo, so that feedback is very helpful. Sounds like removing the exception handler could be counter-productive indeed. 

Changing the log to an error sounds good to me. I'll wait on more feedback from @sqldiablo and incorporate that change (warning->error) into any other changes that come out of the feedback.

sqldiablo:
Hi guys, I apologize for the radio silence lately. Unfortunately, I don't really have time to invest in this right now, or for the foreseeable future. I thought I would, but I've had several things come up suddenly that are taking my time and attention. Please proceed without me at this point.

esev:
No worries @sqldiablo. Thanks for the update. Hope all is well for you!

pywemo/ouimeaux_device/__init__.py
```diff
@@ -208,6 +198,8 @@ def subscription_update(self, _type, _params):
             try:
                 self._state = int(self.parse_basic_state(_params).get("state"))
             except ValueError:
+                LOG.warning("Unexpected BinaryState value `%s` for device %s. "
```
pavoni: Lets log an error rather than a warning (now we know it should have been important!)

Also, on reflection, since this is on the subscription update, I think it would be better to leave `_state` unchanged rather than setting to `0`. There is a decent chance that the state was set currently (we opportunistically set the internal state).

pavoni (changes requested):
Looks good - have suggested a couple of small changes.

Thanks for the contribution (and sorry about the delay getting to this).

esev:
> Thanks for the contribution (and sorry about the delay getting to this).

You're welcome. I'm happy to help. And no worries about the delay. I understand.

pavoni:
Thanks - now merged.

Let's release when your other PRs are also merged - just about to take a look...

pavoni:
Now released as `0.4.45`.

I think you're a HA user - do you want to so the HA release (so you can double check it works with the new switch)?

Or shall I (although I can't test the new switches).

esev:
Thank you @pavoni for the reviews and the new release. I'll sent a PR for HA to update the version in the wemo component later today after testing that everything is still working properly with my devices.

PR #159, opened 2020-07-27
Reliability improvements for SSDP discovery.

esev:
## Description:

- Retry fetching the xml description from the device.
- Always process all received data before exiting. This handle the case where a single device can take too long and prevent other devices from being discovered.
- Eliminate the 0.5 second delay before processing the first device.
- Probe more than twice if time allows.
- Don't start the timer until after the initial SSDP requests are sent.

**Related issue:** https://github.com/home-assistant/core/issues/36771

## Checklist:


pavoni:
This looks like a great improvement - thanks.

PR #166, opened 2020-09-09
Tidy ghost files etc

pavoni:
## Description:
Remove files from old build process that are no longer needed.

## Checklist:


sqldiablo (approved):
I'm fine with this, but we'll need to looking into getting pytest running happily with the new build process now. It's commented-out currently. I have black commented-out because it's misbehaving as well.

pavoni:
I uncommented pytest  - and it worked fine for me both locally and here. Can you try again?

I also stopped everything running twice on a PR. We can check it still runs on a merge to master when we merge.

Black is another story...

sqldiablo:
Your successful build is evidence that it's working now, as that's the issue I was having on the other PR. I'm fine with this PR getting merged in at this point.

pavoni:
Thanks for your help.

Issue #169, opened 2020-10-19
Port detection fails when using DNS hostname


joshuaboniface:
Attempting to connect to a device via hostname with PyWemo appears to fail:

```
$ host basement-mini01
basement-mini01 has address 10.0.0.62
$ pip list | grep pywemo
pywemo (0.4.46)
$ python3
>>> import pywemo
>>> address = "basement-mini01"
>>> port = pywemo.ouimeaux_device.probe_wemo(address)
>>> print(port)
None
>>> address = "10.0.0.62"
>>> port = pywemo.ouimeaux_device.probe_wemo(address)
>>> print(port)
49153
```

Ideally, this should work equally well with a hostname or IP address.

Came to this conclusion after determining that my previous HomeAssistant configuration, using the Wemo module there, which in turn uses PyWemo, stopped working after a recent update. I am using DNS hostnames in the configuration rather than direct IPs, and verified as above. So this did work at some point in the past (I was on a very old install), but no longer does.

joshuaboniface:
Doing some more digging, this actually looks to be a Wemo-side problem. Hitting the device with a hostname returns a 500, while hitting it with an IP returns a 200 as expected. Perhaps PyWemo should attempt to resolve the IP itself, then probe using the actual IP?

joshuaboniface:
>Please note that you need to use ip addresses as shown above, rather than hostnames, otherwise the subscription update logic won't work.

This is very unfortunate, perhaps it should be updated to mention that devices will sometimes fail with hostnames for their own reasons as well.

pavoni:
Interesting catch. Not sure it's a big deal since most people use discovery anyway (although this is unreliable for me).

We should update the docs as you suggest.

PR welcome - both for the documentation - and the code if you feel strongly - although with what you've discovered sounds like there may be other hidden issues!

joshuaboniface:
Indeed, I'll do a PR to update the docs just to make note of it.

I don't use discovery myself since my PyWemo user (HomeAssistant) is on a different network from my WeMo devices, so I specify everything directly by hostname (and I use those because I have 12+ devices to remember!) - I definitely agree this is an unusual configuration for most home users.

[...]

PR #170, opened 2020-10-19
Add mention of WeMo IP-only firmware bug

joshuaboniface:
## Description:

Recent versions of the WeMo firmware (as of yesterday, 2020-09-19) seem
to throw 500 errors when connected to via hostname, but work fine when
connected to by IP. Mention this in the manual documentation in case
anyone else notices this.

**Related issue (if applicable):** References #169
## Checklist:
  - [x] The code change is tested and works locally. **N/A, documentation only**


pavoni:
Thanks for the detective work - and updating this.

PR #172, opened 2020-11-01
Port setup_url_for_address and resolve hostnames to IPs

### LINKED ISSUE #169: Port detection fails when using DNS hostname

joshuaboniface:
## Description:

1. Ports in the `setup_url_for_address` function from the HomeAssistant WeMo component.
2. Fixes #169 by resolving a provided hostname string (non-IP-address `{host}` value) into an IP address before crafting the URL in the `setup_url_for_address` function.

Will require an updated release containing this change before https://github.com/home-assistant/core/pull/42722 can be accepted.

**Related issue (if applicable):** fixes #169

## Checklist:

### DIFF
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -1,9 +1,9 @@
 """Module to discover WeMo devices."""
-import logging
-import requests
-
 from socket import gethostbyname, gaierror
 from ipaddress import ip_address
+
+import logging
+import requests
 
 from . import ssdp
 from .ouimeaux_device.bridge import Bridge
@@ -111,7 +111,8 @@ def setup_url_for_address(host, port):
 
     if is_hostname:
         try:
-            # The {host} must be resolved to an IP address; if this fails, this will throw a socket.gaierror.
+            # The {host} must be resolved to an IP address; if this fails, this will
+            # throw a socket.gaierror.
             host_address = gethostbyname(host)
 
             # Reset {host} to the resolved address.
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -115,7 +115,7 @@ def hostname_lookup(hostname):
         LOG.debug(
             'Could not resolve hostname %s to an IP address.',
             hostname)
-        return None
+        return hostname
 
 
 def setup_url_for_address(host, port):
@@ -128,11 +128,7 @@ def setup_url_for_address(host, port):
         ip_address(host)
     except ValueError:
         # The provided {host} should be treated as a hostname.
-        host_address = hostname_lookup(host)
-        if host_address is not None:
-            host = host_address
-        else:
-            return None
+        host = hostname_lookup(host)
 
     # Automatically determine the port if not provided.
     if not port:
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -126,12 +126,8 @@ def setup_url_for_address(host, port):
         # Attempt to register {host} as an IP address; if this fails ({host} is not an IP address),
         # this will throw a ValueError.
         ip_address(host)
-        is_hostname = False
     except ValueError:
         # The provided {host} should be treated as a hostname.
-        is_hostname = True
-
-    if is_hostname:
         host_address = hostname_lookup(host)
         if host_address is not None:
             host = host_address
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -96,6 +96,7 @@ def device_from_uuid_and_location(uuid, mac, location,
 
     return None
 
+
 def setup_url_for_address(host, port):
     """Determine setup.xml url for a given host and port pair."""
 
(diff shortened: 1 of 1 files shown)


pavoni (changes requested):
Might be neater to drop the lookup into a function , something like
(please excuse un tested code).
```python
def lookup_hostname(host):
    try:
        # The {host} must be resolved to an IP address; if this fails, this will
        # throw a socket.gaierror.
        host_address = gethostbyname(host)
    except gaierror:
        # The {host}-as-hostname did not resolve to an IP address.
        LOG.debug(
            'Could not resolve hostname %s to an IP address.',
            host)
        return host

    # Reset {host} to the resolved address.
    LOG.debug(
        'Resolved hostname %s to IP address %s.',
            host, host_address)
    return(host_address)
```

And then

```python
def setup_url_for_address(host, port):
    """Determine setup.xml url for a given host and port pair."""

    # Force hostnames into IP addresses
    try:
        # Attempt to register {host} as an IP address; if this fails ({host} is not an IP address),
        # this will throw a ValueError.
        ip_address(host)
    except ValueError:
        # The provided {host} should be treated as a hostname.
        host = lookup_hostname(host)

    # Automatically determine the port if not provided.
    if not port:
        port = probe_wemo(host)

    if not port:
        return None

    return "http://%s:%s/setup.xml" % (host, port)
```

joshuaboniface:
@pavoni Pretty much exactly how I implemented it, but with one exception: after resolving and returning, I do another None check; we want to pass a None back to the HASS if the hostname lookup fails, since at that point we know it's both (a) not an IP, and (b) not a resolvable hostname. So there's one more check there before I replace `host`.

joshuaboniface:
Today is not my day for catching every instance of a variable name...

pywemo/discovery.py
```diff
@@ -91,3 +95,54 @@ def device_from_uuid_and_location(uuid, mac, location,
...
+
+def setup_url_for_address(host, port):
+    """Determine setup.xml url for a given host and port pair."""
+
+    # Force hostnames into IP addresses
+    try:
+        # Attempt to register {host} as an IP address; if this fails ({host} is not an IP address),
+        # this will throw a ValueError.
+        ip_address(host)
+        is_hostname = False
```
pavoni: I don't think the `is_hostname` variable is needed - take another look at my suggestion.

pywemo/discovery.py
```diff
@@ -91,3 +95,54 @@ def device_from_uuid_and_location(uuid, mac, location,
...
+    except ValueError:
+        # The provided {host} should be treated as a hostname.
+        is_hostname = True
+
+    if is_hostname:
+        host_address = hostname_lookup(host)
+        if host_address is not None:
+            host = host_address
+        else:
+            return None
```
pavoni: I'm slightly concerned about a use case where a call is made with a host that we don't understand - but would work perfectly well with the underlying library.

So I think if we don't think what they passed is an ip address, but the lookup didn't work - I think the best fallback is just to try what the gave us - rather then returning none.

It might well not work - but it seems better to try doing what they asked - rather than deciding they were wrong ourselves!

This is the logic that was in my suggested code sample.

pywemo/discovery.py
joshuaboniface: Makes sense.

pywemo/discovery.py
joshuaboniface: I think that's fair. I can't think of any situation where this would matter - it's either an IP, or a resolvable hostname, or an invalid hostname which shouldn't be queried - but I'm OK with that. My main concern here was avoiding the bug I was seeing, and this does it since I personally can be assured that the hostnames resolve.

pavoni (approved):
Looks great - thanks for taking the time to do this.

joshuaboniface:
Glad to! Once there's a new release I'll finish up https://github.com/home-assistant/core/pull/42722

pavoni:
Thanks again @joshuaboniface - I've merged https://github.com/pavoni/pywemo/pull/175 which updates the docs - and makes your code accessible outside the library.

I'll release a new version once https://github.com/pavoni/pywemo/pull/176 is merged - which I think is also important for HA.

pavoni:
Now released as `0.5.2 ` (I decided to split the releases).

PR #178, opened 2020-11-14
Reset setup

### LINKED ISSUE #171: Device Factory Reset/Restore
This is more of a question than an issue, but I'm not really sure where else to post it.  I'm curious if it is possible to do a full factory reset using pywemo?  I've looked at my switches and I've played around with some of the functions I found using the "explain" method as well as poked around in the _config base object, but I didn't find anything related to a full reset.  I'm looking for the equivalent of "Factory Restore" in the app which claims to "Erase all settings and returns it to default", per the screen shot on the Belkin support website.  Since this can be done via the phone app, I'm assuming that it can be done via uPnP using pywemo.  Any insights or pointers on how to achieve this?

Spectre5:
## Description:  Add support to reset and setup wemo devices with pywemo

See the related issue for some background information.  I've created a script to reset and setup Wemo devices.  I've found it to work very well for myself throughout many tests, but I only have 5 total Wemo devices (and only 3 unique models).  I'm able to successfully reset and setup each of the devices.  You can also bulk reset all devices and if the user is on Linux and has nmcli installed, then they can also bulk setup many devices at once.  It does this by using nmcli to search for Wemo.* networks to connect to and then connects to them and sets up whatever Wemo it finds.  Without nmcli, the user just needs to manually connect to whatever Wemo.* network they want to setup (and give the name of the device there).  This has worked well for me in my testing.  To setup the Wemo devices, the user also needs to have openssl installed as it is used to encrypt the password when sending it to Wemo (AES support only, not TKIP).  Also included is a "list" command which prints a lot of useful information about all discovered devices (various Get methods).  I've only tested on Linux and I don't have access to macOS nor Windows to test there.

When I first started this reset/setup script, my intention was to make the script for myself (put it in its own repository).  As such, I've written the script outside of the pywemo core library (it is in the scripts directory) and made it install via [tool.poetry.scripts] in the pyproject.toml file.  The script does pass all pylint and flake8 checks when using scripts/build.sh.

This may introduce some pain points:

- The script added two dependencies:
  - click for the cli interface
  - colorlog to improve readability in the terminal
  - I'm not real familiar with poetry, but adding these seemed to make a lot of changes to the poetry.lock file, even to sections of unrelated libraries, but maybe that is expected.
- I've added black as a dev dependency and added 2 changes to the defaults for it in the pyproject.toml file:
    - line length of 79, per PEP8
    - skip string normalization (single quotes are far cleaner and more readable than double quotes)
    - I can remove black if desired, since it appears that not all code in pywemo passes the black --check
- I've added 1 more ignore to the pylintrc file (bad-continuation) as black does it properly and pylint incorrectly flags this
    - Actually, updating pylint to 2.6.0+ would prevent this need as pylint has decided to drop the bad-continuation check altogether, see [here](https://github.com/PyCQA/pylint/pull/3571).  The version currently locked with poetry is pylint 2.4.3, so we could alternatively upgrade that.

If you'd prefer to skip this separate/external script concept (and remove all of those pain points), then I could implement the core functionality as just 2 or 3 new methods in the "Device" class and then move the cli interface I've written to its own separate repository.  If I did this, then I would basically add modified versions of the "wemo_reset", "wemo_setup", and "encrypt_wifi_password_aes128" functions as class methods on the Device class (I think?).  Let me know your thoughts.

I've also added a firmware warning to the README as mentioned in the related issue.

**Related issue (if applicable):** fixes #171

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,9 +21,6 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Update pip
-        run: |
-          python -m pip install --upgrade pip
       - name: Build
         run: |
           ./scripts/build.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,9 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: |
+          python -m pip install --upgrade pip
       - name: Build
         run: |
           ./scripts/build.sh
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ Dependencies
 ------------
 pyWeMo depends on Python packages: requests, ifaddr and six
 
-provided_setup_reset.py additionally depends on: click, colorlog
+provided wemo_reset_setup.py additionally depends on: click, colorlog
 
 How to use
 ----------
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,6 @@ How to use
 
     >> devices[0].toggle()
 
-
 If discovery doesn't work on your network
 -----------------------------------------
 On some networks discovery doesn't work reliably, in that case if you can find the ip address of your Wemo device you can use the following code.
@@ -38,7 +37,16 @@ On some networks discovery doesn't work reliably, in that case if you can find t
 
 Please note that `discovery.device_from_description` call requires a `url` with an IP address, rather than a hostnames. This is needed for the subscription update logic to work properly. In addition recent versions of the WeMo firmware may not accept connections from hostnames, and will return a 500 error.
 
-The `setup_url_for_address` function will lookup a hostname and provide a suitable `url` with an IP addesss.
+The `setup_url_for_address` function will lookup a hostname and provide a suitable `url` with an IP address.
+
+Firmware Warning
+----------------
+Starting in May of 2020, Belkin started requiring users to create an account and login to the app (Android app version 1.25).
+In addition to the account, most of the app functionality now requires a connection to the cloud (internet access), even for simple actions such and toggling a switch.
+All of the commands that go through the cloud are encrypted and cannot be easily inspected.
+This raises the possibility that Belkin could, in the future, update Wemo device firmware and make breaking API changes that can not longer be deciphered.
+If this happens, pywemo may no longer function on that device.
+It would be prudent to upgrade firmware cautiously and preferably only after confirming that breaking API changes have not been introduced.
 
 Developing
 ----------
--- a/README.rst
+++ b/README.rst
@@ -8,6 +8,8 @@ Dependencies
 ------------
 pyWeMo depends on Python packages: requests, ifaddr and six
 
+provided_setup_reset.py additionally depends on: click, colorlog
+
 How to use
 ----------
 
--- a/README.rst
+++ b/README.rst
@@ -42,7 +42,7 @@ The `setup_url_for_address` function will lookup a hostname and provide a suitab
 Firmware Warning
 ----------------
 Starting in May of 2020, Belkin started requiring users to create an account and login to the app (Android app version 1.25).
-In addition to the account, most of the app functionality now requires a connection to the cloud (internet access), even for simple actions such and toggling a switch.
+In addition to the account, most of the app functionality now requires a connection to the cloud (internet access), even for simple actions such as toggling a switch.
 All of the commands that go through the cloud are encrypted and cannot be easily inspected.
 This raises the possibility that Belkin could, in the future, update Wemo device firmware and make breaking API changes that can not longer be deciphered.
 If this happens, pywemo may no longer function on that device.
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,120 +1,170 @@
 [[package]]
+name = "appdirs"
+version = "1.4.4"
+description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
 category = "dev"
-description = "An abstract syntax tree for Python with inference support."
+optional = false
+python-versions = "*"
+
+[[package]]
 name = "astroid"
+version = "2.3.3"
+description = "An abstract syntax tree for Python with inference support."
+category = "dev"
 optional = false
 python-versions = ">=3.5.*"
-version = "2.3.3"
 
 [package.dependencies]
 lazy-object-proxy = ">=1.4.0,<1.5.0"
 six = ">=1.12,<2.0"
+typed-ast = {version = ">=1.4.0,<1.5", markers = "implementation_name == \"cpython\" and python_version < \"3.8\""}
 wrapt = ">=1.11.0,<1.12.0"
 
-[package.dependencies.typed-ast]
-python = "<3.8"
-version = ">=1.4.0,<1.5"
-
 [[package]]
-category = "dev"
-description = "Atomic file writes."
 name = "atomicwrites"
+version = "1.4.0"
+description = "Atomic file writes."
+category = "dev"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-version = "1.4.0"
 
 [[package]]
-category = "dev"
-description = "Classes Without Boilerplate"
 name = "attrs"
+version = "20.2.0"
+description = "Classes Without Boilerplate"
+category = "dev"
 optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
-version = "20.2.0"
 
 [package.extras]
-dev = ["coverage (>=5.0.2)", "hypothesis", "pympler", "pytest (>=4.3.0)", "six", "zope.interface", "sphinx", "sphinx-rtd-theme", "pre-commit"]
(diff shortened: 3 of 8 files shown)


Spectre5:
I had to update pip in the virtual environment to make the build work correctly in python 3.6, otherwise I had a similar error to what is mentioned [here](https://github.com/python-poetry/poetry/issues/2688).

pavoni:
Wow - this is pretty amazing. 

Of the extra libraries `colorlog` is so tiny that it isn't an issue `click` is a bit heavier - but there is quite a bit of interaction here - so I can see that it's needed.

I managed to get it to reset and then setup a  UK `WeMo_WW_2.00.11483.PVT-OWRT-Insight` Insight Switch and an `WeMo_WW_2.00.11408.PVT-OWRT-SNS` Switch with no issue.

I was able to reset a `WeMo_WW_2.00.11423.PVT-OWRT-Maker` Maker - but had some issues with setting it up again. I rebooted the Wemo and then on the second attempt it worked (though I was going to have to create an account!)
```
poetry run wemo_reset_setup setup --ssid XXX   --name="XXX" --password XXX
[INFO    ] starting discovery...this may take a few seconds
[WARNING ] Error fetching description at http://10.22.22.1:49152/setup.xml
[WARNING ] Error fetching description at http://10.22.22.1:49152/setup.xml
[WARNING ] Error fetching description at http://10.22.22.1:49152/setup.xml
[CRITICAL] device named "XXX" not found

[greg@Gregs-MacBookPro16] ~/development/ha/pywemo (reset-setup) ⚡
❯ poetry run wemo_reset_setup setup --ssid XXX   --name="XXX" --password XXX
[INFO    ] starting discovery...this may take a few seconds
[CRITICAL] device named "XXX" not found

[greg@Gregs-MacBookPro16] ~/development/ha/pywemo (reset-setup) ⚡
❯ poetry run wemo_reset_setup list
[INFO    ] found 0 devices
```

Rebooted the WeMo
```
❯ poetry run wemo_reset_setup setup --ssid XXX   --name="XXX" --password XXX                                                                                                                                                                  
[INFO    ] starting discovery...this may take a few seconds
[INFO    ] found device with name: XXX
[INFO    ] found AP with SSID: XXX
[INFO    ] using this access point data: XXX|11|81|WPA2PSK/AES
[INFO    ] sending connection request (try 1)
[INFO    ] pairing status: Connecting
[INFO    ] starting status checks (20 second timeout)
[WARNING ] Error communicating with XXX at 10.22.22.1:49152, retry 0
[ERROR   ] Unable to re-probe wemo at 10.22.22.1
[INFO    ] Trying to reconnect with XXX
[INFO    ] Hi Fi Systemline Sensor Not found in try 0. Trying again in 0 seconds
  File "/Users/greg/development/ha/pywemo/.venv/lib/python3.8/site-packages/urllib3/connection.py", line 169, in _new_conn
    conn = connection.create_connection(
  File "/Users/greg/development/ha/pywemo/.venv/lib/python3.8/site-packages/urllib3/util/connection.py", line 96, in create_connection
    raise err
  File "/Users/greg/development/ha/pywemo/.venv/lib/python3.8/site-packages/urllib3/util/connection.py", line 86, in create_connection
    sock.connect(sa)
OSError: [Errno 51] Network is unreachable
```
and then finally a setup with no issue.


I'm on a mac - so the `nmcli` option didn't work - but gave a sensible error message.

I found a few minor issues where the prompt text didn't agree with the code (`all` vs `setup-all` and `-v` being suggested with the setup option to list APs which didn't seem to work).

So I think the real question is whether to include it here - or set-up another repo as you suggest (with the core function added here)

I think over time more people will object to the requirement for a network account - so it's really nice to have the option to avoid that. On the other hand pywemo is very stable - and not changing much - and it feels like this might need a few revisions for device compatibility and stability. 

What are your feelings about support etc?

The command line docs are very good - but if we include here we should add a paragraph or two about how to use the script.

If you'd still like to include and support here I'll go through the PR code more carefully.

Thanks again - I think this is an amazing piece of work!

pavoni:
BTW I also had to run `poetry lock` to get this to work, which then made a lot of changes to the lock file.

I wonder if we're using different poetry versions?

```
❯ poetry --version
Poetry version 1.0.10
```

If so very happy to upgrade (and should probably work out how to lock this!)

Also had to switch to `black = "^20.8b1"` - couldn't find `b2`

Spectre5:
I'm glad to see that it has worked for someone else!  A handful of notes based on your experience:

- I agree that click is pretty heavy.  I could implement this script using argparse, but by habit I usually reach for click first as I find its interface a bit nicer to work with and I always have it already installed.
- Even when using the Wemo app, I've often had to attempt setting up a device 2 or 3 times before it worked, so I'm not surprised that you had some issues.  In my first version, device setup would fail nearly every time the first time, and then would work on the second attempt.  There is a printed log message or two in the code regarding this, to make sure users are aware that they should attempt setup again if it fails.  When trying to debug and understand this, I found that sending the ConnectHomeNetwork command twice would result in my device connecting on the "first" try almost every time, which is why I added this part:

    ```python
    # success rate is *much* higher if the ConnectHomeNetwork command is
    # sent twice (not sure why!)
    for _ in range(2):
        result = device.WiFiSetup.ConnectHomeNetwork(
            ssid=ssid,
            auth=auth_mode,
            password=encrypted_password,
            encrypt=encryption_method,
            channel=channel,
        )
        time.sleep(0.15)
    ```

- The discovery code is straight from pywemo (pywemo.discover_devices), so there is nothing unique in the script there.  So I'm not sure why you had some trouble finding it at first.  In the second case, it appears that pywemo lost the device for some reason.  I'm not sure why, but I did experience that once or twice during my initial tests as well.  However, before submitting this PR, I was able to reset and setup all devices 3 or 4 times in a row without issue, so I was hopeful that some of my edits made it more stable.  The warning about communication and error about re-probing comes from pywemo (ouimeaux).  So it seems that it got "lost" after the ConnectHomeNetwork command was sent?
- As I mentioned, I only have Linux and so used nmcli.  I image there is a macOS equivalent that could be used and if you were interested in playing with it, then we could add support for both and select the appropriate one based on the OS.  I believe that Windows could use `netsh` to do this as well, but I didn't attempt to add add it since I cannot test it.
- Good catch on the reference to `--all`; that should be `--setup-all` and I've fixed that now.  As for `-v`, this could probably be re-written to make it more clear, but it is implemented in the code.  Basically, the `-v/--verbose` command applies to all of the cli "commands" and so, the way click works, it needs to be applied *before* the `setup` command is given.  So, for example, this should work (`-v` before `setup`):
`poetry run wemo_reset_setup -v setup ...other options...`
- It looks like I have poetry version 1.1.4 installed in the virtual environment.  From the build.sh script, it just calls `pip install poetry`, so the latest version of poetry will be installed when someone runs that.  In general that is probably fine, but I guess if you want to control that too then you could use `pip install poetry==1.1.4` to control the poetry version.  Then you control the poetry version the same way that poetry controls the versions of libraries that it installs.
- Weird about the black library.  In my branch, the installed version of black is `20.8b1` and that is the version referenced in the repositories `pyproject.toml` (`black = "^20.8b1"`) and `poetry.lock` files.  Where did you need to change `b2` to `b1`?
- I'm happy to provide whatever support I can, but as I'm sure you've found with pywemo already, it can be quite hard to provide support when you cannot duplicate a users environment/device/firmware/etc.

So, where does that leave us?  I'm happy to contribute this code if you want and it would probably have better visibility in pywemo.  However, I think pywemo is really more of a library to connect and interface with a device than it is to actually control devices.  So it may be better to add the reset and setup functionality to pywemo and then move this "setup" script to a separate repository, which would also remove the heavy click dependency (although I'll miss having the automated pywemo builds!).  I'm happy to take a stab at what this would look like, if we want to compare.  I think we'd want to add these new methods to the `Device` class, which is in the `ouimeaux` directories `__init__.py` file.  Is that a file that I should edit, or does pywemo try to avoid editing the `ouimeaux` directory?

pavoni:
Thanks for the thoughtful reply.

Let's aim for a separate repo (for the reason you highlight) - but keep the code closely linked. Perhaps call it pywemo_setup or something similar.

I'm happy to host a repo (and give you co-ownership) if that helps keep them closely associated  - but equally happy to include some links to a repo you own.

The automatic build code was kindly contributed to one of the other repos I maintain, and I've just replicated it to the others, very happy to do the same again. I'm sure we can get this running easily. You probably don't even need the automatic pypi/build release section.

Spectre5:
Sounds good.  Would you still want to add a "setup" and "connect" method to the Device object in pywemo, or keep that stuff only in this separate/new script as well?  If you do want that part in pywemo, then would you want me to edit the Device class in `ouimeaux/__init__.py`?

pavoni:
If you’re willing I think it makes sense to add those methods to the `Device` class.

Spectre5:
Ya, I agree that the `Device` class is the best place.  I just wasn't sure exactly what your relationship with `ouimeaux` is (if any at all anymore).  In other words, I wasn't sure if try to avoid making changes to that directory so you could pull in changes from it easily.

Spectre5:
I'll put together a new PR with the smaller change for pywemo when I find the time, and then we can figure out where to put the separate script.

pavoni:
The `ouimeaux` code that was forked from https://github.com/iancmcc/ouimeaux has barely changed over the years.

But that’s more credit to the the author than any policy.

So no issue making changes if you wish.

Spectre5:
Closing this pull request in support of splitting the supplied script into methods in pywemo and a separate script to use the methods as discussed above.

Issue #181, opened 2020-12-11
Discusson: poetry.lock file


Spectre5:
This is just meant as a quick discussion.  Why does pywemo commit the poetry.lock file?  I understand that it provides deterministic builds, but I think it is meant more for applications (final products), less so for libraries.  I see pywemo as a library.

In general, poetry does suggest that you commit it [in their docs](https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control), however it also indicates in that section that "For libraries it is not necessary to commit the lock file."

pavoni:
I'm certainly not an expert. And even now it isn't even working the way I'd expect (the recent release built fine on my local machine - and failed on github until I updated the various library versions).

This probably means something is missing in poetry.lock.

My main aim is to make sure the build on github is the same as my local build - and I think the lock file would help, rather than hinder this!

Views from experts very welcome!

Spectre5:
Ya, I'm no expert with poetry either.  That said, my understanding is that if the poetry.lock file is present, poetry should use it exactly (which is your goal here), and if it is not present, then it uses what is in the pyproject.toml file.  If they differ, then it complains (gives a warning).  If you lock the pyproject.toml file to an exact version (use == instead of ^ or another variation of versioning), then it would use those exact versions anyways, even without the lock file.  I think that in general though, libraries don't usually try to lock to really specific versions and usually want to be as "open" as possible to different versions.

I just did a test --- I cloned this library to a new directory and ran the `build.sh` script, then activated the environment.  The poetry version is 1.1.4 (it uses `pip install poetry` in `build.sh`, so the latest version will be put into the virtual environment).  Then if I simply run `poetry lock`, it results in a different lock file (may lines changed)!  However, it appears that the file is essentially the same, but that this newer version just orders it differently.  The newer one is more reasonable, really, since the name and version are listed first.  As a quick example, here is the first package in each:

Current master:

    [[package]]
    category = "dev"
    description = "Atomic file writes."
    name = "atomicwrites"
    optional = false
    python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
    version = "1.4.0"

updated with `poetry lock`:

    [[package]]
    name = "atomicwrites"
    version = "1.4.0"
    description = "Atomic file writes."
    category = "dev"
    optional = false
    python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"

[...]

PR #182, opened 2020-12-13
Add black and isort

Spectre5:
## Description:  Add black and isort, bump pylint

Let me first say that this looks like a much bigger change that it is.  In terms of the code itself, there should effectively be no change at all.  I saw from build.sh that black is "temporarily" disabled, so I took a shot and enabling it, along with isort.  I also updated pylint to 2.6.0 since they removed the bad-continuation with that version.  You could break down this commit into the following:

- Add black section to pyproject.toml as well as black and isort dependencies (for dev version) and bumping pylint.
- Edits configs files so that pylint, flake8, and black are all enabled for all files except those in the `xsd` directory.  This does make flake8 and pylint take a little longer since they check more files, but it isn't too bad.  All of them expect the PEP8 recommended 79 characters per line.
- Edit the build.sh to enable support for isort and black
- Edit pylintrc to enable some "good names" and make some edits to code comments for those (removed some pylint locally disabled checks)
- Some minor code comment updates to support flake8 and pylint (since they are enabled for more files than previously --- they are now enabled on the `ouimeaux_device` directory, just not the lower `xsd` directory).
- The poetry.lock file has been updated and re-locked for these new libraries (and there are many other changes in terms of re-ordering the other dependencies as I found and mention in issue #181)

Most of the file changes in this PR were updated entirely by black and isort by running `black .` and `isort .` after they were added with poetry.  My local build is clean and all checks and tests pass.  I've also deleted the poetry.lock file and virtual environment to try testing it "from scratch" and it worked for me too.

**Related issue (if applicable):** N/A

## Checklist:


Spectre5:
Also feel free to just close this PR if you're not interested in adding these

pavoni:
Thanks for taking the time to do this. 

I had a go a while back and had black, pylint and isort fighting each other - but sounds like you resolved this!

Spectre5:
I think so, at least locally and in the github CI.  Have you had a chance to test it and play around with it?  Are you interested in merge this?

pavoni:
All good - works nicely.

Thanks for taking the time to do this.

I've added you as a contributor - hope that's OK.

Issue #177, opened 2020-11-12
Support for Wemo Outdoor Plug


carterthayer:
Discovery and manual input for this device does not work with this device

pavoni:
Do you have any more info? Do you get any error traces?

It's quite possible that Belkin are moving away from their local control model - as suggested here:-

https://github.com/pavoni/pywemo/issues/171#issuecomment-721463048

If that's the case then this pywemo may be unable to control new devices.

carterthayer:
No errors. It was not found when doing discovery and the device object is None when using the IP method from the README.

I will do some more digging to figure out why. Do you have any guidance/recommendations for looking at what is happening with new devices?

pavoni:
I've seen some discussions about Belkin now requiring a cloud account registration - and moving from a local network model of control to a cloud one. So it's possible that the new device doesn't support the old style api.

If the device will talk to you there are parts of the code here can help in getting service names etc. but if you can't even get the device to talk to you - then the they only way is to use something like wireshark to see what is happening on the network between the Wemo app and the device as @spectre5  is doing here https://github.com/pavoni/pywemo/issues/171 but that is not an easy job!

Spectre5:
When the commands go to/from the cloud, they are encrypted so figuring out what the commands are from the app will be quite difficult, if not impossible.  With wireshark you'll be able to see the traffic, but to decrypt that message you'll need the decryption key.  This is possible in some cases where Firefox can save the key to a file for you, as described
[here](https://unit42.paloaltonetworks.com/wireshark-tutorial-decrypting-https-traffic/) or [here](https://www.comparitech.com/net-admin/decrypt-ssl-with-wireshark/).  However, I don't know how you would get the decryption key from the wemo app.  Maybe there is some android app that could intercept and output it for you, but I'm not aware of any.

Basically, I wouldn't be surprised if new and future Wemo devices don't work with pywemo.  With that said, I'd be a little bit surprised if the entire protocol is different for the outdoor plug, since I have a few switches on the latest firmware and pywemo still works fine.  So before you give up on this outdoor plug, you should debug it some more.

- Are you sure you're connect to the same network?
- Can you ping in at the IP?
- Try nmap to see what ports are open on it.
- Fire up wireshark and see what you can see, I imagine that some of the communication between the app and device is still local.

mattbilodeau:
I was able to get details from my outdoor plug, via netdisco:

{'host': '192.168.1.7',
  'mac_address': '{mac}',
[...]

PR #183, opened 2020-12-15
add UnsupportedWemo

Spectre5:
## Description:  Return an ~~UnknownDevice~~ UnsupportedWemo if pywemo doesn't recognize an expected Wemo device

New Wemo products won't yet exist in pywemo and currently discovery would return `None` for those devices.  This PR adds a mechanism to allow a user to manually interact with the product, assuming that it "looks" like other wemo products.  The idea for this was prompted by #177 and I'll re-post a few of those thoughts here.

If a devices does not exist in pywemo, then `device_from_uuid_and_location` returns `None`, so the device is never "discovered".  This PR would add an `UnknownDevice` which is basically the same as the base `Device` class.  This would allow some future Wemo device to still be discovered, just not directly usable since it wouldn't have the helper methods like `on`, `off`, etc defined for it.  But at least it would let a user know that something was discovered on the network.

I'm not sure if there would be downstream effects to this (i.e. in Home Assistant).  Downstream users could though easily ignore devices that return `True` for `isinstance(object, UnknownDevice)`.  Home Assistant may want to do this since the unknown device can't be easily controlled.

**Related issue (if applicable):**  N/A

## Checklist:

### DIFF
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -6,7 +6,7 @@ from socket import gaierror, gethostbyname
 import requests
 
 from . import ssdp
-from .ouimeaux_device import probe_wemo
+from .ouimeaux_device import UnsupportedDevice, probe_wemo
 from .ouimeaux_device.api.xsd import device as deviceParser
 from .ouimeaux_device.bridge import Bridge
 from .ouimeaux_device.coffeemaker import CoffeeMaker
@@ -121,6 +121,19 @@ def device_from_uuid_and_location(
         )
     if uuid.startswith('uuid:OutdoorPlug'):
         return OutdoorPlug(
+            url=location, mac=mac, rediscovery_enabled=rediscovery_enabled
+        )
+    if uuid.startswith('uuid:'):
+        # unsupported device, but if this function was called from
+        # discover_devices then this should be a Belkin product and is probably
+        # a WeMo product without a custom class yet.  So attempt to return a
+        # basic object to allow manual interaction.
+        LOG.info(
+            'Device with %s is not supported by pywemo, returning '
+            'UnsupportedDevice object to allow manual interaction',
+            uuid,
+        )
+        return UnsupportedDevice(
             url=location, mac=mac, rediscovery_enabled=rediscovery_enabled
         )
 
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -701,3 +701,22 @@ class Device:
     def serialnumber(self):
         """Return the serial number of the device."""
         return self._config.get_serialNumber()
+
+
+class UnsupportedDevice(Device):
+    """Representation of a WeMo device without a definition in pywemo.
+
+    This class is used if an apparent WeMo device is found on the network via
+    upnp discovery, but the device does not yet exist in pywemo.  This will
+    allow a user to see that something is discovered and manually interact with
+    it as well as aide in creating a permenant class for the new product.
+    """
+
+    def __repr__(self):
+        """Return a string representation of the device."""
+        return '<WeMo UnsupportedDevice "{name}">'.format(name=self.name)
+
+    @property
+    def device_type(self):
+        """Return what kind of WeMo this device is."""
+        return 'UnsupportedDevice'


Spectre5:
If we're worried about downstream issues due to this, we could also add a `debug` or `use_unknown_device` or similar keyword to `discover_devices` that gets passed to the `device_from_uuid_and_location` method.  Then give it a default of `False` and only if it is `True` would an `UnknownDevice` be returned.

pywemo/discovery.py
```diff
@@ -93,15 +93,27 @@ def device_from_uuid_and_location(uuid, mac, location,
     if uuid.startswith('uuid:Humidifier'):
         return Humidifier(url=location, mac=mac,
                           rediscovery_enabled=rediscovery_enabled)
+    if uuid.startswith('uuid:'):
+        try:
+            # unknown device, try to make a basic object to allow manual
+            # interaction
+            return UnknownDevice(url=location, mac=mac,
```
pavoni: Should this also log something, along with some info about the device (eg `unsupported device X found - returning  Unknown Device`).

pywemo/discovery.py
```diff
@@ -93,15 +93,27 @@ def device_from_uuid_and_location(uuid, mac, location,
...
+                                 rediscovery_enabled=rediscovery_enabled)
+        except (AttributeError, requests.exceptions.RequestException):
```
pavoni: Slightly concerned about this `failing silently`. I recall seeing an error about retrieving `Setup.XML` at some point. It might have been before we checked the device manufacturer - or might have been a crashed WeMo device.

Fine to catch this is you think it's needed - but perhaps put something in the log?

I think by the time we get here - we know it is a Belkin Device - but if that's what you want to check for - perhaps don't return an Unknown device.

pavoni (commented):
I think this is a great idea - a few small suggestions.

pywemo/discovery.py
Spectre5: I've added a short log here, but see further discussion below.

Spectre5:
Let's talk about how to best implement this.  Based on what I said above and what you've noticed in the code, I think we need to resolve the following:

- Should there be a unique `UnknownDevice` returned or simple a base class `Device` object?  We do know it is a Belkin product already from the snippet below in the discovery method.  I though it would make sense to return an object that makes it clear that pywemo doesn't really know about the product, but I'm not set on that if we'd rather just return `Device`.  Maybe it would be more clear to return `UnknownWemo`, `ManualWemo`, or something else like that?

        if entry.match_device_description(
            {'manufacturer': 'Belkin International Inc.'}
        ):

- Should the new device (or basic `Device` class) object always be returned, or only with some sort of `debug` or similar argument set to `True`?  If so, what argument name would you think is most clear?  This probably depend on the point above.

pywemo/discovery.py
Spectre5: see my new comment below, but yes we do know this is a Belkin device.  So calling it "unknown" is perhaps a bit of a misnomer, but I just meant that the product is unknown to pywemo.

I caught those two exceptions just because I didn't want some unfamiliar product to cause issues with the continued discovery - perhaps a future product will have a different setup.xml schema.  The `AttributeError` can be raised if a setup.xml file is found but doesn't contain the information needed (for example if it didn't include `friendlyName`).  But I'm happy to just let it raise an exception instead too.

pavoni:
Agree with you comments about `Unknown`, but I like your approach to split it out as a new class - because it makes it very clear to an API user that there is a device that we don't support. Perhaps just change the classname to `UnsupportedWemoDevice` or something similar?

As long as we log an error I'm fine with you catching those exceptions, but don't like to catch exceptions like that silently unless needed. Spent several days once once finding a bug where a library author was catching something he probably shouldn't have been - and doing it silently!

I see two classes of users we need to tell a device is unsupported. The API users are perfectly handled by your `UnknownDevice` class (whatever you decide to call it).  End users who have software that uses the API are probably best served by a clear error log - just so they aren't surprised that their device isn't supported.

In HA we'll probably trap the `UnknownDevice` and give an error anyway - but I still think it's good for the library to clearly report a warning / error.

Spectre5:
Ok, I've removed the exception catching and renamed it to `UnsupportedWemo`.  Still happy to rename it again to whatever makes the most sense.  I also considered `UnsupportedDevice` since it could potentially not be a "Wemo" product...

This would raise an error if you call `device_from_uuid_and_location` directly with bad data, which is fine.  My only concern would be if an error/exception caused the rest of the discovery to be stopped.  Below are a few examples of failures that could happen, although these examples should not happen from the `discover_devices` function since the URL should be good and the the device should be a Belkin product.  I could see this potentially causing an issue if Belkin releases a product someday that doesn't follow the specification of the rest of the wemo products, which would probably result in an `AttributeError` (or something else?).  But we can probably cross that bridge when the day comes.

Also, these examples would be issues even if the uuid was a known product.  So these are not unique to the new class exactly, but it is a higher chance that a bad request is sent to it since we'd be fairly certain that the recognized wemo products would behave as expected.

- A bad request (capital XML)

    >>> pywemo.discovery.device_from_uuid_and_location('uuid:Unknown', None, r'http://192.168.1.30:49153/setup.XML')
      File "/home/scott/github/pywemo/pywemo/discovery.py", line 136, in device_from_uuid_and_location
        return UnsupportedWemo(
      File "/home/scott/github/pywemo/pywemo/ouimeaux_device/__init__.py", line 118, in __init__
        service_list = self._config.serviceList
    AttributeError: 'NoneType' object has no attribute 'serviceList'

- A bad port

    >>> pywemo.discovery.device_from_uuid_and_location('uuid:Unknown', None, r'http://192.168.1.30:50000/setup.xml')
      File "/home/scott/github/pywemo/.venv/lib/python3.8/site-packages/urllib3/connection.py", line 169, in _new_conn
        conn = connection.create_connection(
      File "/home/scott/github/pywemo/.venv/lib/python3.8/site-packages/urllib3/util/connection.py", line 96, in create_connection
        raise err
      File "/home/scott/github/pywemo/.venv/lib/python3.8/site-packages/urllib3/util/connection.py", line 86, in create_connection
        sock.connect(sa)
    ConnectionRefusedError: [Errno 111] Connection refused
    ...<snip>...
    requests.exceptions.ConnectionError: HTTPConnectionPool(host='192.168.1.30', port=50000): Max retries exceeded with url: /setup.xml (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f05e4bd3070>: Failed to establish a new connection: [Errno 111] Connection refused'))

- A bad IP address results in a timeout ConnectionError, of course

Spectre5:
For what it is worth, this is my one worry with not trying to catch exceptions, "That said, I feel like one bad wemo shouldn't bring down the whole platform?" from [this HA issue](https://github.com/home-assistant/core/issues/23744).  Another option might be to add a `failsafe` boolean to the discovery function that would log and ignore certain exceptions like `AttributeError`, xml exceptions, and maybe requests exceptions.  Don't want to make it too broad of course, as that is never a good thing either, but if it is trying to be `failsafe` (default false), then I assume that we'd want to be relatively broad.

The solution in that issue was to unplug and re-plugin the device, I don't think we can help with that!  And that might be a case where swallowing the exception and ignoring the device would be bad...

pavoni:
Thanks for this - now merged and released as `0.5.6`

Do you want to do the HA update - or shall I?

Spectre5:
If you have the time to do it, that'd be great.  But I'd like to see the PR you create (for my own understanding on the HA PR process).  If you can't do it right now, then I can take a stab at it in a couple days, probably.

pavoni:
HA PR here https://github.com/home-assistant/core/pull/44476

PR #185, opened 2020-12-15
Add readme validation

pavoni:
## Description:
This adds validation to the ReadMe file that has caused problems when releasing.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ python = "^3.7 || ^3.9"
 ifaddr = ">=0.1.0"
 requests = ">=2.0"
 six = ">=1.10.0"
+rstcheck = "^3.3.1"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -17,6 +17,7 @@ python = "^3.7 || ^3.9"
 ifaddr = ">=0.1.0"
 requests = ">=2.0"
 six = ">=1.10.0"
+rstcheck = "^3.3.1"
```
Spectre5, line 20: @pavoni I think this should be in `[tool.poetry.dev-dependencies]` section of the pyproject.toml?

pavoni: Thanks https://github.com/pavoni/pywemo/pull/187

PR #186, opened 2020-12-16
update generateDS generated code

Spectre5:
## Description:  Update the `device.py` and `service.py` files in the xsd directory.

The current code in the `xsd` directory was `Generated Thu Jan 31 15:50:44 2013 by generateDS.py version 2.8b` and it appears to have been done with python 2.X.  I found that when using the `export` method from a device that binary strings were being output to the file as literal binary strings, which is likely to do with the differences in python 2 and 3 strings.  I think it makes sense to update these files with a newer python version.  This PR uses `generateDS.py version 2.37.10` from my `python 3.8.5` installation.

These files are really at the heart of pywemo as they create the entire underlying communication protocol.  This makes it even more important to update it, but it also means that this should be well tested to ensure that this doesn't have secondary effects that get missed.  I've tested it with my few devices and it seems to work for me.

A few more notes:

- I've added the exact version of generateDS used to the dev dependencies for posterity
- I've added lxml to the dependencies.  lxml is used by default with generateDS, but if it is not found then it falls back to the inbuilt python `xml.etree`.  However, I found that this causes issue with an attribute access to `prefix` on lines such as `self.ns_prefix_ = node.prefix`.  This isn't unique:
  - [This issue](https://github.com/sync-or-swim/pyfixm/issues/4) for pyfixm solved it by adding lxml as a dependency.
  - [This issue](https://github.com/andrebrait/1g1r-romset-generator/issues/3) for another library solved it by added a dummy method to return `None` if the attribute doesn't exist, as shown in [this commit](https://github.com/andrebrait/1g1r-romset-generator/commit/402ef60534ae4e25bd64eb066caf0f93dde84373).  I tested this in our device and service python files and it *did* fix the issue and allowed the removal of `lxml` as a dependency, but I feel like using `lxml` is a safer solution and would ensure that the processing would be the same for all users.  Otherwise some people may have `lxml` installed as a dependency from another library and others may not.
- The parseString method signature has been changed along the way.  The old/current signature is `parseString(inString)` while the new/updated one is `parseString(inString, silence=False, print_warnings=True)`.  I've thus added these two extra keywords to all parseString calls in pywemo since the default is `silence=False` which prints a bunch of information to stdout.  I've used `silence=True` and `print_warnings=False` in all calls to `parseString`.

**Related issue (if applicable):** N/A

## Checklist:


pavoni:
This seems like a good idea - lets merge and release the other things - and then take a bit more time to test this.

The library used to work with Python2 - although not sure it still does.

If this requires Python3 - we should perhaps make it a major version change.

Spectre5:
Ya, based on your README update with removing python 2 from the description, I assumed that python 2 support was effectively dropped.  Plus it isn't included in the CI testing.  Anyway, although it is completely untested by me, it appears that these updated files do intend to support python 2 still anyway:

    if sys.version_info.major == 2:
        BaseStrType_ = basestring
    else:
        BaseStrType_ = str

But I certainly agree that this needs more testing.  I'll update this PR again to resolve conflicts after the others get merged.

pavoni:
Just to say that I'm fine to merge this once you've rebased / sorted out the conflicts.

Spectre5:
Updated, should be good now.  I updated generateDS.py to 2.37.11, although there is no effective different in the device.py and service.py files that it generates.

pavoni:
~I got an error when I tried to run this (the previous version of this PR was fine)~
```
❯ python
Python 3.8.5 (default, Jul 21 2020, 18:31:18)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pywemo
>>> devices = pywemo.discover_devices()
  File "/Users/greg/development/ha/pywemo/pywemo/discovery.py", line 48, in discover_devices
    device = device_from_description(
  File "/Users/greg/development/ha/pywemo/pywemo/discovery.py", line 63, in device_from_description
    parsed = deviceParser.parseString(
  File "/Users/greg/development/ha/pywemo/pywemo/ouimeaux_device/api/xsd/device.py", line 2358, in parseString
    rootObj.build(rootNode, gds_collector_=gds_collector)
  File "/Users/greg/development/ha/pywemo/pywemo/ouimeaux_device/api/xsd/device.py", line 1097, in build
    self.ns_prefix_ = node.prefix
AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'prefix'
>>>
```

~Can do some digging if the problem isn't obvious to you.~

Must have been using an old library somewhere.

Was fine when I re-installed.

pavoni:
Thanks - now merged.

BTW now you have access to the repo - you can create a dev branch directly - rather than using your fork.

Might make your workflow a little easier.

Spectre5:
Sure, will do.  Just FYI about that error you saw - see the second note in the original PR post.  Basically that error exists if lxml is not installed, which is why that was added as a dependency.

PR #187, opened 2020-12-16
Move rstcheck to be a dev dependency.

pavoni:
## Description:
I incorrectly added this as a library dependency

## Checklist:


Stefan41199:
4878d62b7544109b47c4ad1e27006b1ffcdd58e6

Issue #61, opened 2017-01-27
Wemo Light Switch Long Press


dlamkins:
I could be wrong, but based on my understanding of the eventservice.xml file hosted by the WeMo lightswitches, it looks like the WeMo Light Switches now support the long press event via UPnP. My switches are running firmware version "WeMo_WW_2.00.10937.PVT-OWRT-LS" which I got on Jan 25, 2017. Here is my eventservice.xml contents:

[http://pastebin.com/zJgesnAc](http://pastebin.com/zJgesnAc)

I've highlighted the two sections regarding long presses. The first looks to be for simulating a long press on the WeMo and the second appears to be an event that can be subscribed to. Here are those sections pulled out:

**Action (In Action List)**

```
<action>
<name>SimulateLongPress</name>
<argumentList>
<argument>
<name>LongPress</name>
<direction>out</direction>
<relatedStateVariable>LongPress</relatedStateVariable>
</argument>
</argumentList>
</action>
```

**Event (In Service State Table)**

```
<stateVariable sendEvents="yes">
<name>LongPress</name>
<dataType>string</dataType>
<defaultValue>0</defaultValue>
</stateVariable>
```

Is it possible that this feature could get added?  I would really like to eventually be able to take advantage of this feature in Home-Assistant.

Thanks!

pavoni:
Pull requests welcome. 

Afraid I don't have a Light Switch myself - so not much I can do.

TaylorTWBrown:
If I donate a switch, can you build it? :D

pavoni:
Afraid I'm based in the UK - and wemo don't support / supply the bridges or light switches here.

Don't know if @jaharkes would be open to help.

jwkicklighter:
@pavoni I'm taking a look at this, but having trouble following how this library works. In `subscribe.py`, I see nothing that limits the type of events that are listened to. In theory, if the Wemo Switch is sending long press events shouldn't the logger catch that the event was received?

pavoni:
@jwkicklighter it's easy to miss - take a look here:-
https://github.com/pavoni/pywemo/blob/master/pywemo/subscribe.py#L113

You can see the subscription to the `basicevent` service - and the commented out subscription to the `insightevent` service. The current code doesn't handle multiple subscriptions for a single device - hence commenting out the second subscription.

The HA wrapper refreshes the insight switch - so we get the state changes without the subscription (for now).

[...]

PR #190, opened 2020-12-22
Add Python classes for interacting with the rules database

esev:
## Description:

Adding a `RulesDb` class and some python ORM helpers for interacting with the on-device rules database. I've included enough capability to add/remove a virtual long-press device and to generate the database from scratch if it does not already exist. See https://github.com/pavoni/pywemo/compare/master...esev:long_press (still a work-in-progress) for the larger context about where this will fit.

To test this PR with a device:

```
import logging
import pywemo
from pywemo.ouimeaux_device.api.rules import rules_db_from_device

WEMO_IP = ''  # Fill this in

device = pywemo.discovery.device_from_description(
    pywemo.discovery.setup_url_for_address(WEMO_IP, 0), None
)

with rules_db_from_device(device) as device_rules:
    print(device_rules.rules)
    print(device_rules.rule_devices)
    print(device_rules.target_devices)
    print(device_rules.rules_for_device())
```

## Checklist:


Spectre5:
Just FYI - from your git directory, running `./scripts/build.sh` will run `black` and `isort` on the code and report on the any flake8/pylint linters.  If the code passes running `./scripts/build.sh', then it should pass the CI check used for the PR

esev:
I think there may be some pylint issues with 3.9. Should I remove the type annotations from this PR?
https://github.com/PyCQA/pylint/issues/3882#issuecomment-745148724

Spectre5:
Just seeing this now.  Sorry there was some issues with the builds, happy to look at suggestions or improvements there.  As for type annotations, I'd personally prefer to add them but since none of the code has them so far, I've not added them with the few PRs that I've made.  At some point I think it would be good to go through and add type annotations to the code base though!

Spectre5:
But looks like you did get there anyway.  @pavoni, have you had a chance to look into this work yet?  I know it can be a busy time of year though!

pavoni:
I was going to let you guys reach the end of the discussion (and to fix the lint issues) before looking in detail!

The detective work here is amazing!

I was slightly concerned abut adding a database library - although I hadn't realised that sqlite is now part of python3 - so 
that should be fine.

I think you are both HA users - which would cover my other concern - which is we need to make sure the library continues to work well inside HA.

My current thoughts are that we merge all three of the outstanding PRs - and then do a major release - since they are all fairly significant changes.

esev:
Just to provide some background on my thought process on adding support for long press events: I wanted to start with this PR, on its own, separate from the changes to `ssdp.py` and `subscribe.py`, as the rules database is something new, and I wanted to get feedback on whether or not this API was appropriate. I had started with something very simple (https://github.com/esev/pywemo/commit/34f08c9f66f0a724e9c8460bee989cb7359de0cf), but then realized there was a lot more information in the rules database than just long press (scheduling, etc). I thought it would be good to setup some structured API for accessing the database that can be easily extended in the future if someone wants to work on adding support for the other database entities.

> I was slightly concerned abut adding a database library - although I hadn't realised that sqlite is not part of python3 - so
> that should be fine.

I shared this same concern and was also happy to discover that sqlite3 is a built-in library. I wanted to use a ORM library, like `peewee`, but wasn't sure if it was worth adding an additional dependency for. That's why I added a few classes to do this internally. If you feel like this was the wrong decision, and I should use something like `peewee`, let me know and I'll change things around.

esev:
@Spectre5 it seems `isort` & `black` are having a dispute. :)  Can you offer a pointer on how to address this?

```
$isort pywemo/ouimeaux_device/api/rules.py
Fixing pywemo/ouimeaux_device/api/rules.py

$ git diff
diff --git a/pywemo/ouimeaux_device/api/rules.py b/pywemo/ouimeaux_device/api/rules.py
index c6069ae..fd22ca5 100644
--- a/pywemo/ouimeaux_device/api/rules.py
+++ b/pywemo/ouimeaux_device/api/rules.py
@@ -8,16 +8,7 @@ import sqlite3
 import tempfile
 import zipfile
 from types import MappingProxyType
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    FrozenSet,
-    List,
-    Mapping,
-    Optional,
-    Tuple,
-)
+from typing import Any, Callable, Dict, FrozenSet, List, Mapping, Optional, Tuple
 
 import requests
 
$ black pywemo/ouimeaux_device/api/rules.py
reformatted pywemo/ouimeaux_device/api/rules.py
All done! ✨ 🍰 ✨
1 file reformatted.

$ git diff
$ 
```

pavoni:
Just to confirm I'm comfortable with `sqlite` because it doesn't add a dependency.

So confirm when you're happy and we can review and merge.

And will be interested to see what @Spectre5 says about the `isort` vs `black` dispute. Previously I couldn't get them to play nicely together - but he seemed to fix it!

esev:
I'm happy with this PR as-is now and will welcome any feedback. Thank you!

Spectre5:
Hey, sorry for the isort/black heartache.  I remember when the two tools worked together to get on the same page so that they worked nicely together and I thought that isort made that black-compatible configuration the default.  I was wrong.  Please update the pyproject.toml file to change:

    [tool.isort]
    profile = "hug"

to:

    [tool.isort]
    profile = "black"

That should make isort fully compatible with black!

Spectre5:
Ok, I've updated the code to use profile "black" for isort configuration in the master code, so you can pull in that change.  I'll look at the PR closer and other comments above when I have a little more time.

esev:
> Ok, I've updated the code to use profile "black" for isort configuration in the master code, so you can pull in that change. I'll look at the PR closer and other comments above when I have a little more time.

Looks like I needed to adjust the line_length for isort too. That seems weird to me though because it looks like 79 is the default.

Spectre5:
Ah yes, you are right.  Because the "black" profile sets the line length to the default 88 used by black, but we've changed that to 79 here in the black configuration.  I should have caught that, thanks!

Spectre5:
It looks like @esev has done a lot of really great work here to bring this all together and he's had a bit of success running it locally!  I definitely want to look at it closer myself soon.  @pavoni, I think you should review this too though, to make sure you're happy with it as this database management portion is entirely new and you'll have better insight than me on potential conflicts or ways to ease integration with home assistant down the line.

As Eric has already said, this is not the full implementation, but basically adds a new rules.py file to interact with the Wemo database(s).

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+
+LOG = logging.getLogger(__name__)
+
+# RulesRow.Type values.
+RULE_TYPE_LONG_PRESS = "Long Press"
+
+VIRTUAL_DEVICE_UDN = "uuid:Socket-1_0-PyWemoVirtualDevice"
+
+
+class DatabaseRow:
```
pavoni: This looks like generic ORM style code (although good code).

I haven't use the Python sqllite bindings -but isn't there a DRY way to do this?

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+        pk_value = self.primary_key_value()
+        sql = f"DELETE FROM {self.TABLE_NAME} WHERE {pk_name}=?"
+        try:
+            cursor.execute(sql, (pk_value,))
+        except sqlite3.Error:
+            LOG.exception("Query failed: %s %s", sql, pk_value)
+            raise
+
+
+class SQLType:
```
pavoni: Same here - looks like code we'd ideally inherit from somewhere.

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+
+    @property
+    def sql_type(self):
+        """Return the sqlite type name for this type."""
+        if self.not_null:
+            return f"{self._sql_type} NOT NULL".lstrip()
+        return self._sql_type.lstrip()
+
+
+class PrimaryKey(SQLType):
```
pavoni: And here.

pavoni (commented):
There's quite a lot of code here that I'm going to trust you guys to test against devices etc.

It all looks well written!

My only real question is that quite a chunk of the code is a mini ORM managing in memory rowsets - and saving and loading them from a general database - with just a small amount of code specifying the database structure - and WeMo specific behaviour.

Should we not be inheriting this type of generic function from a built in library?

I know ORMs can often be overweight - but what's here seems very generic.

esev:
>My only real question is that quite a chunk of the code is a mini ORM managing in memory rowsets - and saving and loading them from a general database - with just a small amount of code specifying the database structure - and WeMo specific behaviour.

> Should we not be inheriting this type of generic function from a built in library?

I'm not able to find any builtin that provides generic ORM features for Python. My goal with rules.py was to abstract away all the DB calls so that it'd be easier to implement features on top (long press, scheduling, auto-off, away mode, etc) and easier to explore the database. However, ORM features are not required in order to implement any of this, and it isn't necessary to hide the database details. 

If you'd prefer, I could remove all the code above the `rules_db_from_device` method and just yield the sqlite3.Connection object from that method. Callers/clients of rules.py can then build directly on-top of the sqlite database. And there wouldn't be so much non-Wemo code to maintain. Is that the direction you'd prefer?

I have a couple other questions:

1. Is there a good place to store a constant that gets shared between the long press logic, ssdp.py, and subscribe.py? Specifically the `VIRTUAL_DEVICE_UDN` is something I'd like to share between all of these. I've put this in rules.py for now, but I'm not sure that is the best spot.
2. Is the `pywemo/ouimeaux_device/api` path the correct location for rules.py? I see there are some other helpers (color.py, util.py) at the top-leve `pywemo` path. Do you have a preference on where this code should go?

Spectre5:
I think that as @esev has already found, there isn't really a built-in ORM codes of sqlite.  There are libraries available, but if we want to stick with pure python3 sqlite, then the ORM portion of the code is probably needed.  It is also why he's put this PR as a separate PR from most of the long-press specific implementation, so that this adds the DB interface code and then other future PR (for long press or other future ideas) would use this interface for the database.

Personally, I'd rather keep the ORM stuff that has been developed here as it is quite clean and clear code and using the ORM interface would be a bit easier than the connection object.  I think the only other idea would be an additional requirement which esev has intentionally avoided here to avoid any extra dependencies.

As for these last 2 questions:

1. I think the constant is probably better kept somewhere else so that this rules.py file is really just the database interface only.  Then the constant(s) would be somewhere else, maybe even in a separate new file?
2. As the path indicates, all code from the`ouimeaux_device` directory was originally taken from the `ouimeaux` project, although there have been some divergence from that code base now.  So the other files like `color.py` and `util.py` were added when pywemo was originally forked and the `ouimeaux` stuff was un-modified.  We could potentially re-organize that directory to make a flatter overall project, but that is certainly out of scope for this PR.  I think the `api` directory makes sense for the rules.py file as it is a lightweight ORM api for the sqlite3 functionality for the project.

Spectre5:
~What about breaking the rules.py into two files?  One that does all the database interaction (lines above the `RulesDb`) and then the more wemo-specific parts being in a separate file (`RulesDb` class and `rules_db_from_device` context manager)?  That would provide a little more separate between the ORM portion of the code and the wemo specific part?~  I'm not sure if this is a good idea.  Obviously the database interaction stuff is also wemo specific based on what rows and fields exist anyway

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+
+@contextlib.contextmanager
+def rules_db_from_device(device) -> RulesDb:
+    """Return a RuleDb instance for the rules on a device.
+
+    The sqlite3.Connection object can be accessed via the '.db' property in the
+    returned RulesDb instance. If the database is modified directly, setting the
+    `.modified` attribute to True will cause the database to be sent to the WeMo
+    device. Any updates that take place via the RulesDb helper methods will also
+    be propogated back to the WeMo device.
```
Spectre5: typo: propogated -> propagated

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+        update(self._target_devices.values())
+        return modified
+
+    def rules_for_device(
+        self,
+        *,
+        device_udn: Optional[str] = None,
+        rule_type: Optional[str] = None,
+    ) -> List[Tuple[RulesRow, RuleDevicesRow]]:
+        """Fetch the current rules for a partular device."""
```
Spectre5: typo: partular -> particular

Spectre5:
I think the code is clean and clear, I have no issues with it and I like the interface.  I've done some testing on it and it does appear to work for me.

- I currently have no rules and that was correctly reported.
- I used the method `ensure_long_press_rule_exists` to add a long press rule and it appears to have been created correctly and re-read correctly.  There is no target since I didn't specify one.
- I also executed `ensure_long_press_rule_exists` on my wemo mini and it was added and re-read without issue as well.  However, the wemo mini does not have the long press functionality, so I assume that the device itself would just ignore that rule.
- For completeness, maybe add a helper function (or method to `RulesDb`) to delete all rules (clear entire database)?  I know the functionality is there to delete a single rule.

pavoni:
@Spectre5 i liked your suggestion about breaking the dB code into two files, why did you change your mind?

In any case I don’t have string feelings about this. The code is nicely written and adds some valuable functionality to the library. 

It is also self contained (at least to this point), and so will have little impact on library uses not needing this.

So happy to merge when you guys are happy.

esev:
Splitting this across three files works if that's what is preferable. We could do something like this:

1. The wemo specific fetch/unpack/pack code (`rules_db_from_device` and the methods it depends on).
2. The database ORM code (maybe including the wemo specific tables).
3. The long press specific code (maybe moving `ensure_long_press_rule_exists` to this file).

The long press module/code can just do something like this:
```
with db_module.rules_db_from_device(device) as conn:
    rules_db = orm_module.RulesDb(conn, device.udn, device.name)
    self.ensure_long_press_rule_exists(db_rules)
```

I think this is a good discussion, and is why I wanted to lead with this PR first. Once we've decided on the structure/API for working with the rules database, I'm hoping the rest should fall into place.

Spectre5:
Right, as is, this has zero impact on current users as this code wouldn't be used at all by HA or anyone not intentionally using it.  So we can also easily adjust this as Eric adds the rest of the long press functionality, if needed.

As for splitting the file, I'm happy either way.  The reason I back tracked was because realistically a lot of the database code written is specifically for the RulesDb anyway since the fields are specific for it.  Maybe a better split would be to but the completely generic classes in a separate file and then all of the rules related subclasses in another?  Just a potential idea

Spectre5:
I thought about that too @esev.  Let's try that.  Three files.  One with totally generic database ORM stuff that could be used with any other database if the need arrives.  Second file with rules database manipulation.  And third file for long press specific code (which will probably be expanded with future PR's).  Does that make sense?

esev:
This makes sense to me. I'll restructure things this evening (in about 7-8 hours).

Thanks for all the feedback. It helps me understand the pywemo project better.

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+            value = f"{value} AUTOINCREMENT"
+        return value.lstrip()
+
+
+class RulesRow(DatabaseRow):
+    """Row schema for the RULES table."""
+
+    TABLE_NAME = "RULES"
+    FIELDS = {
+        "RuleID": PrimaryKey(int, sql_type=""),
```
Spectre5: Is `int` the right type for this `RuleID`?  From [a comment in the reference issue](https://github.com/pavoni/pywemo/issues/61#issuecomment-748693894), the field type is `RuleID PRIMARY KEY` and when inserted it uses `'501'`, as a string?

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+        self.add_rule(new_rule)
+        self.add_rule_devices(
+            RuleDevicesRow(
+                RuleID=new_rule.RuleID,  # pylint: disable=no-member
+                DeviceID=device_udn,
+                GroupID=0,
+                DayID=-1,
+                StartTime=60,
+                RuleDuration=86340,
+                StartAction=ACTION_TOGGLE,
```
Spectre5: Are we missing `EndAction=-1.0,` here?

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+        values = []
+        for name in self.FIELDS.keys():
+            if hasattr(self, name):
+                values.append("%s=%r" % (name, getattr(self, name)))
+        class_name = self.__class__.__name__
+        values_str = ", ".join(values)
+        return f"{class_name}({values_str})"
+
+    def __eq__(self, other):
+        """Test for quality between two instances."""
```
Spectre5: equality

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+        "ZBCapabilityEnd": str,
+        "OnModeOffset": int,
+        "OffModeOffset": int,
+        "CountdownTime": int,
+        "EndTime": int,
+    }
+
+
+class TargetDevicesRow(DatabaseRow):
+    """Row schema for the RULEDEVICES table."""
```
Spectre5: TARGETDEVICES table

pywemo/ouimeaux_device/api/rules.py
```diff
@@ -0,0 +1,695 @@
...
+    FIELDS = {
+        "TargetDevicesPK": PrimaryKey(int, auto_increment=True),
+        "RuleID": int,
+        "DeviceID": str,
+        "DeviceIndex": int,
+    }
+
+
+class DeviceCombinationRow(DatabaseRow):
+    """Row schema for the RULEDEVICES table."""
```
Spectre5: DEVICECOMBINATION table

pywemo/ouimeaux_device/api/rules.py
esev: Fixed

pywemo/ouimeaux_device/api/rules.py
esev: Fixed

pywemo/ouimeaux_device/api/rules.py
esev: We sure are. Good catch. Fixed.

pywemo/ouimeaux_device/api/rules.py
esev: Nice catch. This is interesting. RuleID seems as-if it is meant to match with the other RuleID values in the other tables (which are all INTEGER values in the database). From a Python standpoint, it's important that these all be the same type so comparisons work correctly (RulesRow.RuleID == RuleDevicesRow.RuleID should match). The database seems to be handling the INTEGER/TEXT type juggling just fine. However if there was ever a non-integer TEXT value in the database that would cause issues for pywemo.

As long as the other RuleID fields are INTEGERs, I think it's safe to assume there will be an int value in the RULES.RuleID field. So, I think it's okay to keep as an int within pywemo. The alternative could be to convert all the RuleID values to a str type in pywemo and keep their sql_type as INTEGER. What do you think?

esev: https://www.sqlite.org/faq.html#q3

> Data of any type can (usually) be inserted into any column. You can put arbitrary length strings into integer columns, floating point numbers in boolean columns, or dates in character columns.

Ooof! I didn't realize this. That explains why it is okay to store `NOSYNC` in an INTEGER column. If the types of the fields are not guaranteed, then I'm feeling less confident that the ORM library is a good idea now. If we're already seeing cases where string values are in integer fields, it stands to reason that other integer fields could become strings in the future. That could end up breaking pywemo due to the type juggling in the ORM library and increasing the cost to maintain that code.

I'm starting to think the right thing to do is to not implement that ORM layer and do field comparisons within the database; either with JOIN statements or with passing fetched values back into the WHERE clauses. That'll avoid future maintenance issues due to the wrong type being assumed in the ORM Row classes.

What do you think?

Spectre5: Huh.  I'm not particularly experienced with SQL, so this certainly surprises me!  It does also indicate that using the integer PK though is 1 case where it does matter!  Weird to me that Wemo didn't decide to use an integer for that PK indicated above.  And those docs also indicate that an attempt will be made to convert it if a type is provided....so if `'501'` was provided to an integer column, then it should be converted to `501`.

As for your idea of removing the ORM layer...I'm not sure that it would "fix" this case indicated here?  If we compared the string `'501'` with integer `501`, then it wouldn't match anyway?  I guess Belkin must be doing a conversion within their code to look it up?

I don't suspect that Belkin would be motivated to change the types of their data with any sort of frequency, so that seems pretty low risk to me.  As you found though, they can put whatever they want in the column anyway.  Long story short, I'm not sure exactly what this code would look like without the ORM layer and I'm probably not experienced enough with SQL to make the judgement on which way is better here.  I'd need to fall back on your judgement there.

Spectre5:
I don't see the old comment about the missing `EndAction` anymore, but why did the code even work when that was missing?  I think that adding an `INSERT` in SQL without the right number of columns defined should raise an error.  Or am I wrong there?

esev:
> I don't see the old comment about the missing `EndAction` anymore, but why did the code even work when that was missing? I think that adding an `INSERT` in SQL without the right number of columns defined should raise an error. Or am I wrong there?

The ORM code is doing something like this:

```
INSERT OR REPLACE INTO RULEDEVICES (RuleDevicePK, RuleID, DeviceID, [remaining known fields]) VALUES(1, 1, 'udn');
```

That's why the missing `EndAction` was not causing a database error. It wasn't in the list of known fields that comes after the table name. The database was just inserting a `NULL` in place of the field since it wasn't part of the INSERT statement.

pywemo/ouimeaux_device/api/rules.py
esev: > I don't suspect that Belkin would be motivated to change the types of their data with any sort of frequency, so that seems pretty low risk to me.

I think you're right. Let's go with what we have now. We can revisit in the future if there turns out to be changes.

pywemo/ouimeaux_device/api/rules.py
esev: Fixed in https://github.com/pavoni/pywemo/pull/190/commits/75e48bc36a987e1328eb30920fd6d11e82ffd13b

pywemo/ouimeaux_device/api/rules.py
esev: Fixed in https://github.com/pavoni/pywemo/pull/190/commits/75e48bc36a987e1328eb30920fd6d11e82ffd13b

pywemo/ouimeaux_device/api/rules.py
esev: Fixed in https://github.com/pavoni/pywemo/pull/190/commits/75e48bc36a987e1328eb30920fd6d11e82ffd13b

esev:
> For completeness, maybe add a helper function (or method to RulesDb) to delete all rules (clear entire database)? I know the functionality is there to delete a single rule.

Added `RulesRb.clear_all` in https://github.com/pavoni/pywemo/pull/190/commits/71313fcb2a6bf89b6ad4d93362721436f9642d5b

Spectre5:
Ok, so are we happy to merge this now @esev and @pavoni?  I think the last outstanding thing we had right now was momentarily re-considering the ORM portion of the code as mentioned [here](https://github.com/pavoni/pywemo/pull/190#discussion_r550610946).  For now, we can leave it as is and revisit it or change it in the future.  Is there anything else you're planning to change with this PR at this time?

esev:
> Ok, so are we happy to merge this now @esev and @pavoni?

> Is there anything else you're planning to change with this PR at this time?

I'm happy with the PR as it stands now. I'm not planning on making any more changes.

pavoni:
Merged. Thanks for all your hard work and investigation here!

And thanks to @Spectre5 for the detailed review.

I’d guess there is no benefit to doing a release at this point, although happy to do so if it helps.

PR #192, opened 2021-01-01
Support a virtual device to receive long press events

esev:
## Description:

Extend the `subscribe.RequestHandler` class to support acting as a virtual Wemo device. This allows a physical Wemo device to be able to send long press notifications to pywemo.

## Checklist:

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -180,6 +180,7 @@ class RequestHandler(BaseHTTPRequestHandler):
             self.send_response(200)
             self.send_header("CONTENT-LENGTH", "0")
             self.send_header("TIMEOUT", "Second-1801")
+            # Using a randomly generated valid UUID (uuid.uuid4()).
             self.send_header(
                 "SID", "uuid:a74b23d5-34b9-4f71-9f87-bed24353f304"
             )
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -72,6 +72,17 @@ def get_ip_address(host='1.2.3.4'):
         del sock
 
 
+def _start_server():
+    """Find a valid open port and start the HTTP server."""
+    for i in range(0, 128):
+        port = 8989 + i
+        try:
+            return HTTPServer(('', port), RequestHandler)
+        except (OSError, socket.error):
+            continue
+    return None
+
+
 class RequestHandler(BaseHTTPRequestHandler):
     """Handles subscription responses and long press actions from devices.
 
@@ -221,7 +232,6 @@ class SubscriptionRegistry:
 
         self._http_thread = None
         self._httpd = None
-        self._port = None
 
     @property
     def port(self) -> int:
@@ -271,7 +281,7 @@ class SubscriptionRegistry:
             host = get_ip_address(host=device.host)
             headers.update(
                 {
-                    "CALLBACK": '<http://%s:%d>' % (host, self._port),
+                    "CALLBACK": '<http://%s:%d>' % (host, self.port),
                     "NT": "upnp:event",
                 }
             )
@@ -346,22 +356,10 @@ class SubscriptionRegistry:
         """Add an event callback for a device."""
         self._callbacks[device.serialnumber].append((type_filter, callback))
 
-    def _find_port(self):
-        """Find a valid open port to run the HTTP server on."""
-        for i in range(0, 128):
-            port = 8989 + i
-            try:
-                self._httpd = HTTPServer(('', port), RequestHandler)
-                self._port = port
-                break
-            except (OSError, socket.error):
-                continue
-
     def start(self):
         """Start the subscription registry."""
-        self._port = None
-        self._find_port()
-        if self._port is None:
+        self._httpd = _start_server()
+        if self._httpd is None:
             raise SubscriptionRegistryFailed(
                 'Unable to bind a port for listening'
             )
@@ -407,7 +405,7 @@ class SubscriptionRegistry:
         """Start the HTTP server."""
         self._httpd.allow_reuse_address = True
         self._httpd.outer = self
-        LOG.info("Listening on port %d", self._port)
+        LOG.info("Listening on port %d", self.port)
         self._httpd.serve_forever()
 
     def _run_event_loop(self):
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -169,7 +169,9 @@ class RequestHandler(BaseHTTPRequestHandler):
             self.send_response(200)
             self.send_header("CONTENT-LENGTH", "0")
             self.send_header("TIMEOUT", "Second-1801")
-            self.send_header("SID", "uuid:virtual-device-uuid")
+            self.send_header(
+                "SID", "uuid:a74b23d5-34b9-4f71-9f87-bed24353f304"
+            )
             self.send_header('Connection', 'close')
             self.end_headers()
         else:
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -85,16 +85,16 @@ class RequestHandler(BaseHTTPRequestHandler):
       rules configured for the device. A long press rule is activated whenever
       the button on the Wemo device is pressed for 2 seconds. The long press
       rule is meant to be used to control the state of another device (turn
-      on/off/toggle). However for pywemo's use, a long press rule can be used to
-      trigger an event notification. This is implemented by configuring the
+      on/off/toggle). However for pywemo's use, a long press rule can be used
+      to trigger an event notification. This is implemented by configuring the
       Wemo device to "control the state" of a virtual Wemo device. The virtual
       device is implemented by this class.
 
       The do_GET/do_POST/do_SUBSCRIBE methods below implement a virtual Wemo
       device. The virtual device receives requests to change its state from
       other Wemo devices on the network. When a Wemo device is configured to
-      change the state of the virutal device via a long press rule the following
-      sequence occurs:
+      change the state of the virtual device via a long press rule the
+      following sequence occurs:
 
       1. The Wemo device will attempt to locate the virtual device on the
       network. This is handled by the pywemo.ssdp.DiscoveryResponder class. See
@@ -108,14 +108,14 @@ class RequestHandler(BaseHTTPRequestHandler):
       notifications, but this step seems to be necessary before the next step
       can happen. This step is implemented by the do_SUBSCRIBE method.
 
-      4. When a person presses the button on the Wemo for 2 seconds a long press
-      rule is triggered. If the long press rule is configured with an action
-      for the virtual device, the Wemo device will then call the do_POST method
-      to update the BinaryState of the virtual device. This doesn't actually
-      update any state, rather the virtual device then delivers the event
-      notification to any event listeners configured to receive events from the
-      pywemo SubscriptionRegistry. The event type for a long press action is
-      EVENT_TYPE_LONG_PRESS.
+      4. When a person presses the button on the Wemo for 2 seconds a long
+      press rule is triggered. If the long press rule is configured with an
+      action for the virtual device, the Wemo device will then call the do_POST
+      method to update the BinaryState of the virtual device. This doesn't
+      actually update any state, rather the virtual device then delivers the
+      event notification to any event listeners configured to receive events
+      from the pywemo SubscriptionRegistry. The event type for a long press
+      action is EVENT_TYPE_LONG_PRESS.
     """
 
     def do_NOTIFY(self):  # pylint: disable=invalid-name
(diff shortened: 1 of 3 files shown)


pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
+      rule is meant to be used to control the state of another device (turn
+      on/off/toggle). However for pywemo's use, a long press rule can be used to
+      trigger an event notification. This is implemented by configuring the
+      Wemo device to "control the state" of a virtual Wemo device. The virtual
+      device is implemented by this class.
+
+      The do_GET/do_POST/do_SUBSCRIBE methods below implement a virtual Wemo
+      device. The virtual device receives requests to change its state from
+      other Wemo devices on the network. When a Wemo device is configured to
+      change the state of the virutal device via a long press rule the following
```
Spectre5: typo on "virtual" here.  Also, a bit of a nitpick here but the configuration of black/isort have the line length set to 79 (the 80th character being the newline), but it looks like you have your right "rule" in your editor set at 80 characters as some lines in the comments go 1 extra character, including the "g" on "following" in this line.

pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
+      rule is triggered. If the long press rule is configured with an action
+      for the virtual device, the Wemo device will then call the do_POST method
+      to update the BinaryState of the virtual device. This doesn't actually
+      update any state, rather the virtual device then delivers the event
+      notification to any event listeners configured to receive events from the
+      pywemo SubscriptionRegistry. The event type for a long press action is
+      EVENT_TYPE_LONG_PRESS.
+    """
+
+    def do_NOTIFY(self):  # pylint: disable=invalid-name
```
Spectre5, line 132: Remove the local pylint disable invalid-name here as you've added it to the pylintrc file (I agree it should be there)

pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
                 for property_ in list(propnode):
                     text = property_.text
                     outer.event(device, property_.tag, text)
 
-        self.send_response(200)
-        self.send_header('Content-Type', 'text/html')
-        self.send_header('Content-Length', len(SUCCESS))
+        self._send_response(200, RESPONSE_SUCCESS)
+
+    def do_GET(self):  # pylint: disable=invalid-name
```
Spectre5, line 148: I'd probably just go ahead and add all of these "do_UPPERCASE" names to the pylintrc file ignore list instead of locally disabling them here.

pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
-        self.wfile.write(SUCCESS.encode("UTF-8"))
+        if body:
+            self.wfile.write(body.encode("UTF-8"))
+
+    def _get_xml_from_http_body(self):
+        """Build the cElementTree root from the body of the http request."""
+        content_len = int(self.headers.get('content-length', 0))
+        data = self.rfile.read(content_len)
+        # trim garbage from end, if any
+        data = data.decode("UTF-8").strip()
```
Spectre5, line 206: The previous version used `data.decode("UTF-8").split('\n\n')[0]` here.  Using strip is potentially different, but I think strip is the correct thing to do here.  It also removes any whitespace at the front, but I think we'd want to, if there was any there anyway.

pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
+        self._send_response(200, RESPONSE_SUCCESS)
+
+    def do_GET(self):  # pylint: disable=invalid-name
+        """Handle GET requests for a Virtual WeMo device."""
+        if self.path.endswith("/setup.xml"):
+            self._send_response(
+                200, VIRTUAL_SETUP_XML, content_type="text/xml"
+            )
+        else:
+            self._send_response(404, RESPONSE_NOT_FOUND)
```
Spectre5, line 155: When the `eventservice.xml` is requested from the virtual device, it will return 404, right?  So the firmware on the actual wemo devices won't get the "known" actions from the `basicevent1` service then.  Since you've tested this and it works, I guess you know that it doesn't matter.  We could potentially return dummy XML data to the physical device to declare the `SetBinaryState(BinaryState, Duration, EndAction, UDN, CountdownEndTime, deviceCurrentTime)` action.  A little more management, but maybe it would be safer in case some wemo firmware versions do care.  What do you think?  Worth it or just cross that bridge if some devices have issues?

pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
+
+    def do_SUBSCRIBE(self):  # pylint: disable=invalid-name
+        """Handle SUBSCRIBE requests for a Virtual WeMo device."""
+        if self.path.endswith("/upnp/event/basicevent1"):
+            self.send_response(200)
+            self.send_header("CONTENT-LENGTH", "0")
+            self.send_header("TIMEOUT", "Second-1801")
+            self.send_header("SID", "uuid:virtual-device-uuid")
```
Spectre5: Should this UUID here not be `VIRTUAL_DEVICE_UDN` which is currently defined as `uuid:Socket-1_0-PyWemoVirtualDevice`?

pywemo/subscribe.py
```diff
@@ -35,32 +73,124 @@ def get_ip_address(host='1.2.3.4'):
...
+                    text = binary_state.text
+                    outer.event(device, EVENT_TYPE_LONG_PRESS, text)
+            self._send_response(200, RESPONSE_SUCCESS)
+        else:
+            self._send_response(404, RESPONSE_NOT_FOUND)
+
+    def do_SUBSCRIBE(self):  # pylint: disable=invalid-name
+        """Handle SUBSCRIBE requests for a Virtual WeMo device."""
+        if self.path.endswith("/upnp/event/basicevent1"):
+            self.send_response(200)
```
Spectre5, line 180: Not particularly important, but you could modify the `_send_response` function to accept an optional dictionary (which are ordered since python 3.7) of data to send after the content-length.  Then you could reuse that function here by using an empty string for the body and passing something like `extra_headers={'TIMEOUT': 'Second-1801', 'SID': 'uuid:virtual-device-uuid'}` for that new keyword.

pywemo/subscribe.py
```diff
@@ -91,6 +221,11 @@ def sleep(secs):
         self._httpd = None
         self._port = None
 
+    @property
+    def port(self) -> int:
```
Spectre5, line 238: I like having this property, but this class does already have the `self._port` variable, right?  Should it be removed in favor of this?

tests/test_subscribe.py
```diff
@@ -0,0 +1,159 @@
+"""Tests for the subscription registry and http server module."""
```
Spectre5, line 1: I haven't looked closely at the tests yet - but it does pass for me as well locally.

pywemo/subscribe.py
Spectre5: Ah, I see I looked at this change backwards.  You've removed `do_NOTIFY` and added this local disable here.  I'm fine either way then.  Perhaps the local disable makes sense since it is only defined once.  Then just use the global list for names that may be used more frequently.  But I'm okay either way --- adding them to pylintrc or just doing it locally.

pywemo/subscribe.py
Spectre5: See updated (second) comment on the previous line regarding do_NOTIFY.

pywemo/subscribe.py
esev: >  Perhaps the local disable makes sense since it is only defined once. Then just use the global list for names that may be used more frequently.

This sounds good to me.

When I added the three new methods, I was surprised by the fact that pylint was complaining about them. It took me some time to figure out why pylint complained about `do_GET`, `do_SUBSCRIBE` & `do_POST` but not `do_NOTIFY`. I'd err on the side of putting fewer things inside pylintrc to avoid surprising behavior.

pywemo/subscribe.py
esev: Wemo uses `\r\n\r\n` at the end of the POST messages, that's why I switched to `.strip()` here. I also think it's okay to remove whitespace at the front, but I could use `.rstrip()` if you'd prefer.

pywemo/subscribe.py
esev: This should technically actually be a UUID (ex `uuid:a74b23d5-34b9-4f71-9f87-bed24353f304`). Wemo doesn't seem to mind, but maybe we should use a valid one. I'll make one up.

```
>>> import uuid
>>> uuid.uuid4()
UUID('a74b23d5-34b9-4f71-9f87-bed24353f304')
```

pywemo/subscribe.py
esev: I'm erring toward giving the Wemo less data. The more data that gets provided to the device, the more chances there are of needing to change the data in the future.

pywemo/subscribe.py
esev: Good call. I removed `_port` and restructured things a bit.

pywemo/subscribe.py
esev: Fixed both the typo and the line length issue.

pywemo/subscribe.py
esev: I could go either way (keep it as-is or modify `_send_response` to take the headers as a parameter). But Wemo doesn't respond with a content-type header, so there might end up being a few more edge cases to handle in `_send_response` if we went that route. If it's okay, let's keep it as-is for now. That way the special-casing is in just the `do_SUBSCRIBE` method. 

```
HTTP/1.1 200 OK
DATE: Mon, 04 Jan 2021 22:26:58 GMT
CONTENT-LENGTH: 0
X-User-Agent: redsonic
SID: uuid:fa79d9d4-1dd1-11b2-9603-8f01388fd38f
TIMEOUT: Second-1801

```

pywemo/subscribe.py
Spectre5: I think keep strip

pywemo/subscribe.py
Spectre5: Right.  I guess I'm just a little surprised that it worked, since technically the wemo doesn't know that the `SetBinaryState` action even exists at all!  But I guess they assume it would, since they'd expect the receiver to be a Wemo anyway.  I'm ok to leave this as-is, especially if it works.  But we should file this away in the back of our brains if there are weird issues that some users report

pywemo/subscribe.py
Spectre5: Sounds good.

pywemo/subscribe.py
```diff
@@ -34,33 +72,138 @@ def get_ip_address(host='1.2.3.4'):
...
+            self._send_response(404, RESPONSE_NOT_FOUND)
+
+    def do_SUBSCRIBE(self):  # pylint: disable=invalid-name
+        """Handle SUBSCRIBE requests for a Virtual WeMo device."""
+        if self.path.endswith("/upnp/event/basicevent1"):
+            self.send_response(200)
+            self.send_header("CONTENT-LENGTH", "0")
+            self.send_header("TIMEOUT", "Second-1801")
+            self.send_header(
+                "SID", "uuid:a74b23d5-34b9-4f71-9f87-bed24353f304"
```
Spectre5, line 185: Maybe add a comment here or at the top of this method about a random valid UUID being selected just to avoid any potential confusion here in the future

esev: Added.

By the way, I added your `build.sh` script as a pre-commit git hook. That's been super helpful in catching mistakes before they are committed.

Spectre5: Ya, I've been thinking about having the build.sh file install a pre-commit git hook.  Or maybe separate the file into 2 parts, one that installs the libraries, poetry, and the pre-commit hook and a second that simply runs the various checks (which is the one that would be run for the pre-commit hook).

Spectre5: added a new issue to act as a reminder

Spectre5:
@pavoni, take a look at these few long press related pull requests when you have a chance, to act as another check on the structure of it all.  I have not tested them on my actual wemo devices yet, but I do intend to.

pavoni:
All looks good. Feel free to merge.

Since I badly broke the subscription code a few years back I feel we should test this carefully once merged (although since you've been running this in HA anyway I expect you've done this).

Spectre5:
Yes, this is why I wanted to get your "okay" first before I merged these, since I know you've mentioned that you had a bad time at one point with a breaking change.  As you mentioned though, Eric has been using it even in HA without issue, so that is really promising.  After we merge all 3 of these, I'm happy to test the functionality and his HA PR too to help ensure a smooth transition.

PR #193, opened 2021-01-01
Inform Wemo devices of the pywemo virtual Wemo device

esev:
## Description:

Add a `DiscoveryResponder` class to Inform Wemo devices of the pywemo virtual Wemo device.

## Checklist:


pywemo/ssdp.py
```diff
@@ -9,17 +9,47 @@
...
 
 MIN_TIME_BETWEEN_SCANS = timedelta(seconds=59)
 
+MULTICAST_GROUP = "239.255.255.250"
+MULTICAST_PORT = 1900
+
 # Wemo specific urn:
 ST = "urn:Belkin:service:basicevent:1"
 
+SSDP_REPLY = f"""HTTP/1.1 200 OK
```
Spectre5, line 29: I'm assuming this is fine, but I do see from the wireshark data I have that the SSDP data seems to all end with windows `\r\n` line endings...

pywemo/ssdp.py
```diff
@@ -213,7 +243,7 @@ def build_ssdp_request(ssdp_st, ssdp_mx):
             'ST: {}'.format(ssdp_st),
             'MX: {:d}'.format(ssdp_mx),
             'MAN: "ssdp:discover"',
-            'HOST: 239.255.255.250:1900',
+            'HOST: {}:{}'.format(MULTICAST_GROUP, MULTICAST_PORT),
```
Spectre5: I think f-strings would be cleaner here - as well as the couple unchanged lines above.  They probably used format before mainly due to python 2.7 support previously.

pywemo/ssdp.py
```diff
@@ -336,16 +366,137 @@ def scan(
...
+
+    The DiscoveryResponder informs Wemo devices of the /setup.xml URL for the
+    pywemo virtual Wemo device. The virtual device is used for receiving long
+    press actions from Wemo devices and is integrated into the
+    SubscriptionRegistry HTTP server.
+
+    Wemo devices are informed of the pywemo virtual Wemo device in two ways:
+
+    1. Wemo devices periodically send UPnP M-SEARCH discovery requests for the
+    to locate other devices on the network. DiscoveryResponder responds to these
```
Spectre5: Same comment as the other PR where some of the comments, like this one, go to 80 instead of 79 characters.

pywemo/ssdp.py
```diff
@@ -336,16 +366,137 @@ def scan(
...
+        """Create a server that will respond to WeMo discovery requests.
+
+        Args:
+            callback_port: The port for the SubscriptionRegistry HTTP server.
+        """
+        self.callback_port = callback_port
+        self._thread = None
+        self._exit = threading.Event()
+        self._thread_exception = None
+        self._notify_enabled = True  # For tests.
```
Spectre5: What is this comment here for, `For tests.`?  Is the notify only needed for test code?

pywemo/util.py
```diff
@@ -54,3 +55,15 @@ def interface_addresses():
...
+def get_ip_address(host='1.2.3.4'):
+    """Return IP from hostname or IP."""
+    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    try:
+        sock.connect((host, 9))
+        return sock.getsockname()[0]
+    except socket.error:
+        return None
+    finally:
+        del sock
```
Spectre5, line 69: I know this function was just moved and isn't new, but should this call `sock.close` before it is deleted?  Or are there cases would calling close would raise another error here?

tests/test_ssdp.py
```diff
@@ -0,0 +1,190 @@
+import queue
```
Spectre5, line 1: Tests not closely looked at for this PR yet either.  I do love that you're adding tests for all of the newly added code.

pywemo/ssdp.py
```diff
@@ -336,16 +366,137 @@ def scan(
...
+            sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+            try:
+                sock.bind((addr, 0))
+                sock.sendto(
+                    (SSDP_NOTIFY % callback_addr).encode("UTF-8"), ssdp_target
+                )
+            except socket.error:
+                pass
+            finally:
+                sock.close()
```
Spectre5, line 423: I have done very little python work with sockets, is there a difference between `sock.close` and deleting it with `del sock`, which is used in the `get_ip_address` function?  Some examples I've seen online call `sock.close()` followed by `del sock`.  It seems like the variable goes out of scope immediately anyway, but I've heard of occasional objects that need an explicit `del` at times.

pywemo/ssdp.py
```diff
@@ -336,16 +366,137 @@ def scan(
...
+                # Periodically send NOTIFY messages.
+                now = datetime.now()
+                if now > next_notify and self._notify_enabled:
+                    next_notify = now + timedelta(minutes=2)
+                    self.send_notify()
+
+                # Check for new discovery requests.
+                if not select.select([sock], [], [], 1)[0]:
+                    continue  # Timeout, no data. Loop again and check for exit.
+                msg, addr = sock.recvfrom(1024)
```
Spectre5: Is 1024 just the right number for the M-SEARCH protocol?

pywemo/ssdp.py
```diff
@@ -336,16 +366,137 @@ def scan(
...
+                if len(lines) < 3 or not lines[0].startswith(
+                    b"M-SEARCH * HTTP"
+                ):
+                    continue
+                if (
+                    EXPECTED_ST_HEADER not in lines
+                    or EXPECTED_MAN_HEADER not in lines
+                ):
+                    continue
+                callback_addr = (get_ip_address(addr[0]), self.callback_port)
```
Spectre5: Isn't `addr` the IP?  Is `addr[0]` just the first digit?

pywemo/ssdp.py
esev: Ah, no. Maybe there is a better way to rephrase that. It's so that I can _disable_ the notify for tests. Outside of the test code, nothing else sets this to `False`. It's easier to test all the socket calls if notify can be tested independently from handling M-SEARCH. I added the comment so someone doesn't later come and look at the code and say "this is never set to False, why is it needed?"

pywemo/util.py
esev: Both `del sock` and `sock.close` end up closing the socket.

**del**: https://github.com/python/cpython/blob/0b858cdd5d114f0890b11b6c4d6559d0ceb468ab/Modules/socketmodule.c#L4993

**close**: https://github.com/python/cpython/blob/0b858cdd5d114f0890b11b6c4d6559d0ceb468ab/Modules/socketmodule.c#L3136

 I think `sock.close` is more explicit about what is intended and that `del sock` is overly verbose (i.e. why not `del` every object when you're done?). Would you prefer one be used over the other going forward?

pywemo/ssdp.py
esev: I used the same value (1024) as is used for discovery. 1500 might be more appropriate for both, but either value will work. None of the M-SEARCH messages I have seen have been that large.

pywemo/ssdp.py
esev: `addr` here is a socket address tuple containing `(address, port)`.  I'll rename it `addr`->`sock_addr`

pywemo/ssdp.py
esev: Fixed.

pywemo/ssdp.py
esev: I changed it to use \r\n style instead, just in case. Good catch.

pywemo/ssdp.py
esev: I switched the code around to use f-strings.

pywemo/util.py
Spectre5: I prefer the `close` method myself, I just wanted to make sure there wasn't anything missed by not actually deleting it since I saw a few examples online as well where they deleted it (per my other comment further below)

pywemo/ssdp.py
Spectre5: ok, right, since this was created with `AF_INET`

pywemo/ssdp.py
Spectre5: makes more sense, thanks

pavoni:
This looks good - feel free to merge.

And thanks for improving the existing code too.

PR #202, opened 2021-01-10
Switch to lxml

### LINKED ISSUE #197: [Discussion] Use lxml.etree instead of cElementTree?
When I upgraded the device.py and service.py files, I added lxml as a dependency, which is already installed on HA as well anyway.  Is it worthwhile to change the `cElementTree` code within the rest of the project to use lxml.etree as well?  lxml is generally a little faster and offers some more advanced features.  With that said, I can't image it would be any noticeable real-world difference with the tiny size of xml files using with Wemo devices and we probably don't need any of the extra features.  The other minor benefit would be pywemo using the same ElementTree implementation as those two underlying files.  I'm still not sure if it is worthwhile though, what do you guys think @pavoni and @esev?

Spectre5:
## Description:
Switch to using lxml instead of the built in python xml cElementTree.  I've tested this locally, but not extensively, so I'm not merging yet.  See the referenced issue for a discussion on the reasoning for this change.

**Related issue (if applicable):** closes #197

## Checklist:


pywemo/subscribe.py
```diff
@@ -188,12 +188,12 @@ def _send_response(self, code, body, *, content_type="text/html"):
...
 
     def _get_xml_from_http_body(self):
-        """Build the cElementTree root from the body of the http request."""
+        """Build the element tree root from the body of the http request."""
         content_len = int(self.headers.get('content-length', 0))
         data = self.rfile.read(content_len)
         # trim garbage from end, if any
-        data = data.decode("UTF-8").strip()
-        return cElementTree.fromstring(data)
+        data = data.decode('UTF-8').strip().encode('UTF-8')
```
esev: data = data.strip()

I think the decode/encode is unnecessary.

esev (commented):
I'm getting an error when trying to discover my devices.

```
>>> devices = pywemo.discovery.discover_devices()
  File "github/esev/pywemo/pywemo/discovery.py", line 48, in discover_devices
    device = device_from_description(
  File "github/esev/pywemo/pywemo/discovery.py", line 75, in device_from_description
    return device_from_uuid_and_location(
  File "github/esev/pywemo/pywemo/discovery.py", line 114, in device_from_uuid_and_location
    return Bridge(
  File "github/esev/pywemo/pywemo/ouimeaux_device/bridge.py", line 46, in __init__
    self.bridge_update()
  File "github/esev/pywemo/pywemo/ouimeaux_device/bridge.py", line 71, in bridge_update
    end_device_list = et.fromstring(end_devices.get('DeviceLists'))
  File "src/lxml/etree.pyx", line 3237, in lxml.etree.fromstring
  File "src/lxml/parser.pxi", line 1891, in lxml.etree._parseMemoryDocument
ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration.
```

esev:
grepping through the code, there could be a few more places where strings need to be converted to bytes before calling et.fromstring.

```
$ grep -rn fromstring pywemo
pywemo/ssdp.py:176:                            tree = et.fromstring(xml)
pywemo/ouimeaux_device/humidifier.py:93:    attributes = et.fromstring(xml_blob)
pywemo/ouimeaux_device/bridge.py:71:            end_device_list = et.fromstring(end_devices.get('DeviceLists'))
pywemo/ouimeaux_device/bridge.py:93:        device_status_list = et.fromstring(status_list.get('DeviceStatusList'))
pywemo/ouimeaux_device/coffeemaker.py:59:    attributes = et.fromstring(xml_blob)
pywemo/ouimeaux_device/maker.py:22:        attributes = et.fromstring(makerresp)
```

I haven't verified there are issues with these calls to et.fromstring, merely pointing out other spots that might cause issues similar to the previous comment about the bridge/discovery.

Edit: humidifier.py, coffeemaker.py & maker.py are probably all safe. They don't have an encoding declaration at the start (ex `<?xml version="1.0" encoding="utf-8"?>`).

pywemo/subscribe.py
esev: Looking at this again, the `.strip()` might be unnecessary too. Maybe lxml is more forgiving about trailing/leading whitespace?

```
>>> from lxml import etree as et
>>> et.fromstring(b'\r\n\t  <a><b>c</b></a>   \r\n\t').tag                                                                                            
'a'
```

esev: Never mind. The `.strip()` is still needed.  Here's another example.

```
from lxml import etree as et
>>> et.fromstring(b'\r\n\t  <?xml version="1.0" encoding="utf-8"?><a><b>c</b></a>   \r\n\t').tag
  File "src/lxml/etree.pyx", line 3237, in lxml.etree.fromstring
  File "src/lxml/parser.pxi", line 1896, in lxml.etree._parseMemoryDocument
  File "src/lxml/parser.pxi", line 1784, in lxml.etree._parseDoc
  File "src/lxml/parser.pxi", line 1141, in lxml.etree._BaseParser._parseDoc
  File "src/lxml/parser.pxi", line 615, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 725, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 654, in lxml.etree._raiseParseError
  File "<string>", line 2
lxml.etree.XMLSyntaxError: XML declaration allowed only at the start of the document, line 2, column 9
```

The `<?xml version="1.0" encoding="utf-8"?>` is what is causing problems for the Bridge device. The result from `end_devices.get('DeviceLists')` includes that encoding declaration at the start. I think the uses of `et.fromstring` in humidifier.py, coffeemaker.py & maker.py are all safe, as they don't include this encoding declaration.

pavoni:
I was about to release - but should I wait for this PR?

Spectre5:
Thanks for looking into this @esev.  @pavoni, let's get this PR fixed up and merged, then release.  I'll try to get to it tonight

pywemo/subscribe.py
Spectre5: Ya, I ran into the encoding be defined in the XML issue as well at one point and thought I fixed it, but I figured there was some issues that probably slipped through.

Spectre5: ok, I removed the decode/encode and just strip it directly now.

Spectre5:
I agree with @esev that the occurrences in humidifier.py, coffeemaker.py, and maker.py should all be safe since there is no embedded encoding and never will be since the string is constructed in pywemo.  But I don't have any of those to actually test it.  I could just encode it, but I guess I'd prefer to not waste a couple cpu cycles if it isn't needed.

In ssdp, I've changed the xml to use .content instead of .text from the requests object, so it is binary now.

As for bridge.py, I guess I'm not sure exactly what those actions are returning (again, I don't have one).  Clearly was is returned must have that embedded encoding though based on what esev saw above.  Please test this and give the PR another double-check if you can, much appreciated.

esev:
@Spectre5 could you rebase this PR to the latest master branch? I added a test that will verify that the bridge works.

esev:
I tested this latest update (51efdab) on my bridge device and can confirm that it is working. I think the new test still needs updating though.

https://github.com/pavoni/pywemo/blob/334250d48eb3072b2dbbb0f92bbf6e6e226d1958/tests/ouimeaux_device/test_bridge.py#L3

Spectre5:
ah ya, forgot that you added that extra test, I've pulled in master

esev (approved):

pavoni:
Thanks very much.

esev:
Small update: I updated my HomeAssistant to include this PR (and everything else since 0.5.6). Everything is working fine.

PR #204, opened 2021-01-11
add dependabot

KnicKnic:
## Description:
Add dependabot file to open up PR's for updated dependencies.

See https://dependabot.com/python/ for more info

## Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    # Check for updates to python dependencies every weekday
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every weekday
+      interval: "daily"      


pavoni:
I haven't used this before - any views @Spectre5?

Spectre5:
I don't have a strong opinion, I'm not familiar at all with dependabot except a quick read about it after this PR came in.

pavoni:
@KnicKnic - Can you explain a bit more about why you'd like us to use this?

Spectre5:
Looks like the config setting is for pip, but pip is only used to install poetry into the virtual environment, then poetry does the management.  Does that part matter?  I see that the docs indicate that it works with poetry.  You can see an example of the PR's this creates in @KnicKnic's branch, FYI.

pavoni:
Looks like it should be OK https://github.com/dependabot/dependabot-core/issues/1925

KnicKnic:
> @KnicKnic - Can you explain a bit more about why you'd like us to use this?

I have a [project](https://github.com/KnicKnic/wemo-auto-kotatsu/) that depends on this project. I had to abandon the previous [wemo library](https://github.com/iancmcc/ouimeaux) I was using due to compatibility issues. My hope was that if this project got regular updates that would not occur with this project.

As far as the pip in dependabot.yml. Yes this is an unfortunate name for dependabot. pip essentially means python, which poetry is one of the package managers that is understood by dependabot.

Dependabot does add load to your project, in the form of extra PR's for dependencies. However these PRs come with easy to view links for the PRs & Release notes that occurred for those dependencies. Most projects view their most recent release as the best version of their product. You end up encountering less known bugs, end up getting better traction on any issues you raise, as well as having more performant code. See example https://github.com/KnicKnic/pywemo/pull/3

I was wondering why this project doesn't use dependabot. I figured its simple enough, rather than ask ill just submit a PR.

Spectre5:
Another question on how it works:  what does it use to determine which version of python it needs to be compatible with?  Pywemo targets python 3.7 - 3.9 for now, so when it updates a dependency, it needs to be sure that it works on all 3 versions.  With poetry there must be a python requirement in the dependencies, so I assume it would use that?  I wonder what it would use for a non-poetry repository.  Obviously the CI for a repository should generally catch a bad version, but I'm guessing that dependabot does something internally to ensure it doesn't make a PR for a non-compatible version.

As for support, I think you can rest assured that pywemo will be maintained for the foreseeable future.  pywemo is the library used by home assistant for interaction with Wemo products, so it is used by thousands of people around the world (total guess, I have no idea what the real number is).  Besides a recent bit of activity to update the library and add some new features, it is mostly in maintenance mode with a focus on adding new Wemo products upon release and an emphasis on stability for the home assistant platform.

Finally, isn't there some benefit to not requiring the latest release of all dependencies?  If you can support all versions newer than some version with the `^` requirement, then the newest is fine, but not required, which makes dependency resolution easier with other libraries?  Granted that not all of the requirements use `^` at the moment --- most probably should switch to that.  And I personally do like to have dependencies on my machine up to date.

KnicKnic:
@Spectre5 i don't know the answers to your questions about how dependabot works. I can say that I have been burned by thinking that all versions will work equally well, and allowing people to run many versions (even though those versions were never tested against my code).

About the benefit of allowing your libraries to match with more versions(lagging most recent) translates to 2 things.
1. A different developer thought there may be breaking changes and isn't comfortable allowing many versions, so they pinned a specific version of it. Why would this be the case for them and not you?
1. A different developer found out a newer version has a breaking change in it, and is disallowing it. This means you may have the same problem so you should also pin or at least test.

If these are the case than these libraries are not as safe to match future versions. We should be launching the upgraded versions into CI.

Lastly if a dependency breaks you. You will catch it quickly, shortening the bug filing time and load on the dependency developer, as their incorrect code will be fresh.

I have not used dependabot for long, started enabling in a few repos I own and going from their. I do not know how impactful it will be to my flow. I figured if I didn't like it, I will turn it back off.

pavoni:
As @Spectre5 says I hope we can continue to support the repo. I've been the main maintainer for some time - and with limited time I've just focused on keeping everything running - mainly because I use wemo for home assistant - and I want to keep the devices in my home working!

There has been a burst of activity recently with @Spectre5 and @esev adding more devices and functionality - which is fantastic.

At this point I don't feel the need to keep all the libraries we use up to date - and don't really want to deal with a string of automatically generated PRs doing this. The library is quite mature - and the current approach seems to have kept up with the very active home assistant repo successfully through many years.

Of course I'd welcome PRs for any library updates that you feel are needed - and am happy to try and help make sure the library works for you.

So I suggest we close this for now, unless @esev or @Spectre5 disagree.

KnicKnic:
Closing per @pavoni

PR #205, opened 2021-01-18
Add tests for the WeMo Zigbee Bridge device

esev:
## Description:

Add tests for the WeMo Zigbee Bridge device.

This PR adds an additional dev dependency on `pytest-vcr`. This allows for recording/playback of the actual responses from the device. I figure since not many people have this device, these tests can be used to help validate changes to pywemo.

Note: This test currently fails with #202 

## Checklist:


pavoni:
Fantastic idea - thanks!

Spectre5:
I've heard of pytest-vcr before, but I haven't used it.  It would be great if we could get real responses from all devices that pywemo supports to use for testing each device!

Spectre5:
To create the vcr YAML files, you just need to use the `@pytest.mark.vcr()` decorator and then it'll actually interact with the device the first time you run it and create the YAML files, and then after that it will just use the YAML file and not try to connect.  And if the YAML file exists and there is a new test added that requests a new URL, then it'll raise an error (default `once` mode).  This seems really cool and easy to use.  We should make "cassettes" for each device.  The YAML files are a little heavy but they'll basically never change.  It is a little bit sad that the content of the YAML files are almost identical too, but obviously they aren't intended to be edited by hand.

Spectre5:
@esev, just FYI that it looks like some "actual data" did slip into these YAML files.  Here are a few examples (may not be exhaustive):

```
<DevUDN>uuid:Bridge-1_0-231438B01000D9</DevUDN>
```

```
<DeviceIDs>0017880108DA898B</DeviceIDs>
```

```
...&gt;0017880108DA898B&lt;...
```

And some friendly names like living room couch light.  I doubt any of it matters as I don't know what anyone would actually do with any of this information, but I wanted to let you know since I know that you've replaced at least some data in the `conftest.py` files `scrub_identifiers` function.

esev:
Thanks @Spectre5. Looks like I forgot that the POST data needs to be scrubbed too. No worries about cleaning this up, it can stay in the files. I think the Bridge has some extra IDs (the Zigbee lights IDs) that other devices don't have, so probably not worth trying to add those to conftest.py. But it might be a good idea to scrub the friendly name for future test cases.

I have these devices. Let me know if you'd like me to add tests and yaml for any of them:

- F7C027 Switch
- F7C028 Motion Sensor
- F7C029 Insight Smart Plug
- F7C030 Smart Light Switch
- F7C031 Link Zigbee Bridge
- F7C059 Wifi Smart Dimmer
- F7C063 Mini Smart Plug
- WLS040 WiFi Smart Light Switch
- WLS0403 Smart Light Switch 3-Way
- WSP080 Wifi Smart Plug
- WSP090 WiFi Smart Outdoor Plug

Spectre5:
I've been reading up some on VCR.  I do have a couple of questions that maybe you could answer?

- It appears that the "body" for the POST does not actually matter.  I tried removing it for 1 of the POST commands and the test still runs fine, see the example below.  So does this mean that VCR does not check the contents of it?  If not, then is there any reason to keep that body string in the YAML?

```
- request:
    body:   <--- removed the string that was here
    headers:
    <snip>
```

- Can VCR use multiple YAML files for the test?  In your 3 YAML files, the top ~90% of the file is the same where it just gets all the services and actions from the device.  I was thinking that each device could have that common part in a separate YAML file and then the "test-specific" part in a second YAML file that would be unique for each test.  Maybe that's more work than it is worth?  It looks like `pytest-recording` may support this, but I haven't tried testing it.
- Is there a clean way to parameterize the tests to run it on multiple cassettes?  For example, we could have multiple switches (different versions or different firmware, if needed) and we might run the exact same set of tests on each of the device "cassettes".  So instead of repeating a function a ton of times, it would be nice to just run the same function against a bunch of cassettes.  I don't immediately see a built-in way to do this, although I guess we could write a decorator to do it easily enough if desired.

esev:
> So does this mean that VCR does not check the contents of it? If not, then is there any reason to keep that body string in the YAML?

Looks like it depends on the [configuration](https://vcrpy.readthedocs.io/en/latest/configuration.html#request-matching). For our current use-case we could probably filter out the body with [before_record_request](https://vcrpy.readthedocs.io/en/latest/advanced.html#custom-request-filtering).

> Can VCR use multiple YAML files for the test? 

Yes. I've been experimenting with this. It's possible to do something like this:

```python
@pytest.mark.vcr()
def test_light_turn_on(vcr):
    with vcr.use_cassette('base.yaml'):
        bridge = Bridge('http://192.168.1.100:49153/setup.xml', '')
    ...
``` 

It is also possible to turn this into a fixture (see #206):

```python
@pytest.fixture
def bridge(vcr):
    with vcr.use_cassette('base.yaml'):
        return Bridge('http://192.168.1.100:49153/setup.xml', '')
```

> Is there a clean way to parameterize the tests to run it on multiple cassettes?

I suspect there is a way to use a combination of the fixture example (above) and a base class for the tests. I'll experiment more with this.

esev:
> Is there a clean way to parameterize the tests to run it on multiple cassettes?

How about something like this?

```python
import pytest

from pywemo import Switch


class Base:

    @pytest.mark.vcr()
    def test_turn_on(self, switch):
        switch.on()

    @pytest.mark.vcr()
    def test_turn_off(self, switch):
        switch.off()


class Test_F7C063(Base):

    @pytest.fixture
    def switch(self, vcr):
        with vcr.use_cassette('WeMo_WW_2.00.11420.PVT-OWRT-SNSV2.yaml'):
            return Switch('http://192.168.1.100:49153/setup.xml', '')


class Test_WSP080(Base):

    @pytest.fixture
    def switch(self, vcr):
        with vcr.use_cassette('WEMO_WW_4.00.20101902.PVT-RTOS-SNSV4.yaml'):
            return Switch('http://192.168.1.100:49153/setup.xml', '')
```

Screenshot 2021-01-22 at 8 02 39 PM

Spectre5:
Looks like I missed that bit in the configuration somehow.  So it does support my findings, since the URL and method is the same, the body contents didn't matter.  At least for POST, I think this makes a lot of sense for our use case since we "know" what we're sending to the wemo device already, and a future change (e.g. newlines change) in the future POST commands would cause issues if we required matching the POST body.

Spectre5:
This seems clear and easily readable to me.  This provides a clean separation of the device firmware and bit of functionality under test.

Spectre5:
> I have these devices. Let me know if you'd like me to add tests and yaml for any of them:
> 
>     * F7C027 Switch
>     * F7C028 Motion Sensor
>     * F7C029 Insight Smart Plug
>     * F7C030 Smart Light Switch
>     * F7C031 Link Zigbee Bridge
>     * F7C059 Wifi Smart Dimmer
>     * F7C063 Mini Smart Plug
>     * WLS040 WiFi Smart Light Switch
>     * WLS0403 Smart Light Switch 3-Way
>     * WSP080 Wifi Smart Plug
>     * WSP090 WiFi Smart Outdoor Plug

Besides looking at the physical device, is there a way to get this model information?  I'm not finding it with the pywemo object.  I get "1.0" for "Model Number" based on what the device returns and I don't see it in any of the other get functions I see.

esev:
I haven't found the model anywhere in the returned responses, unfortunately,

Spectre5:
Ya, doesn't seem that the device even knows its real model number.  Annoying.  Presumably the firmware would be a unique identifier anyway since even if the same firmware was on different devices, it would still behave identically if the firmware really was the same.

PR #206, opened 2021-01-23
Use a fixture to simplify test_bridge.py

esev:
## Description:

Use a fixture to simplify test_bridge.py

**Related issue (if applicable): #205 

## Checklist:


esev:
@Spectre5 WDYT?

Spectre5:
Looks good to me and passes all checks locally and on CI.  I think this is a bit nicer myself, even if it does create an extra manual step when adding new tests/devices (to spit the file).  Happy if I merge it now or are you looking at it closer still?

esev:
> even if it does create an extra manual step when adding new tests/devices (to spit the file).

It actually doesn't. I did this PR manually, but for this (https://github.com/pavoni/pywemo/pull/205#issuecomment-765861254) VCR made the files all automatically.

esev:
> Happy if I merge it now or are you looking at it closer still?

I'm good with this.

Spectre5:
> > even if it does create an extra manual step when adding new tests/devices (to spit the file).
> 
> It actually doesn't. I did this PR manually, but for this ([#205 (comment)](https://github.com/pavoni/pywemo/pull/205#issuecomment-765861254)) VCR made the files all automatically.

Wow, really?  That's nice.  But at least the "firmware" part is still manually, right?  To create this cassette:

`with vcr.use_cassette('WeMo_WW_2.00.11057.PVT-OWRT-Link.yaml'):`

I'm talking about a manual step of creating this firmware cassette/yaml versus the test cassette/yaml file.

esev:
> But at least the "firmware" part is still manually, right? To create this cassette:

All I did was looked at the /setup.xml URL manually first to find the firmware version. Then I named the the string for the use_cassette method to match. VCR took care of recording the requests into that file automatically for the first test case (test_turn_on) and then replayed those requests for the second test case (test_turn_off).

The record/replay for the fixture is independent (and uses a different file) from the test cases.

Spectre5:
Really cool.  I'll play around with this using my devices too!

Spectre5:
It's like magic.  It correctly separates out the two files to the "firmware" file and the "test" file.  Totally awesome.  Adding new firmware/devices becomes trivial then, so long as we can find someone with each device --- I think you have all of my devices, plus some :)

So doing this correctly separated into two files, `test_light_turn_on.yaml` and `WeMo_WW_2.00.11452.PVT-OWRT-SNSV2.yaml`, with the vast majority in the firmware file:

```python
@pytest.fixture
def switch(vcr):
    with vcr.use_cassette('WeMo_WW_2.00.11452.PVT-OWRT-SNSV2.yaml'):
        return Switch('http://192.168.1.100:49153/setup.xml', '')

@pytest.mark.vcr()
def test_light_turn_on(switch):
    switch.on()
    assert switch.get_state() == 1
```

PR #207, opened 2021-01-23
Add integration tests for the F7C027, F7C063 and WSP080 switches

esev:
## Description:

Add integration tests for the F7C027, F7C063 and WSP080 switches.

## Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,10 +33,27 @@ def vcr_config():
             b'<MacAddr>001122334455</MacAddr>',
             body,
         )
+        body = re.sub(
+            b'<friendlyName>[^<]+</friendlyName>',
+            b'<friendlyName>WeMo Device</friendlyName>',
+            body,
+        )
+        body = re.sub(
+            b'<hkSetupCode>[^<]+</hkSetupCode>',
+            b'<hkSetupCode>012-34-567</hkSetupCode>',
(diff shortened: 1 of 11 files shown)


pavoni:
I have a couple of UK Insight Switches, a regular SNS switch and a Maker. They look to be the same model number (except with UK at the end) - so assume that just the different electrical hardware.

So I think we're covered except for the Maker - I'll add a similar test once we've merged this.

pavoni:
I added the Maker tests https://github.com/pavoni/pywemo/pull/208

pavoni:
Looks good to me!

Spectre5:
same!

Spectre5:
@esev - I know I previously advocated for removing the body from the cassettes, but now I'm wondering to myself if that was the wrong move and if it would make more sense to instead actually explicitly check the body as part of the test.  I mentioned that "we know what we sent the device", but if we check that what pywemo sends compared to the cassette then it adds some assurance that we still send the same thing in the future too (in case that part of the code changes or refactors at some point).  The main issue I see would be if we intentionally changed the signature (i.e. spacing or something), then we'd just need to make sure we updated all of the cassettes to match the new signature as part of the same PR.

esev:
@Spectre5 - sounds good. See #217

PR #209, opened 2021-01-25
Support multiple subscriptions for the Insight device

### LINKED ISSUE #115: Allow pywemo to subscribe to multiple services
At present the pywemo subscription code only subscribes to the `basicevent` services.

This updates on the most important events like (on / off) - but doesn't work for everything (eg for wemo insight switches - it doesn't update when the power consumption is changed (unless it also triggers a change in or out of standby - which does trigger a `basicevent`).

For this reason I had to leave polling turned on in Home Assistant for switches (@sqldiablo - you might want to check this still works in your recent HA refactor) so that insight parameters would still be updated correctly.

Insight parameter changes are reported by a subscription to `insight`. I expect a number of other wemo devices that report their state via other services such as  https://github.com/pavoni/pywemo/issues/6).

The ghost of my attempt to update this is here
https://github.com/pavoni/pywemo/blob/master/pywemo/subscribe.py#L119

The current re-subscribe code is set up to handle a single subscription per device code so will need to be refactored to support this.

cc @sqldiablo in case you are interested in taking a look at this.

esev:
## Description:

Support multiple subscriptions for the Insight device.

**Related issue (if applicable):** fixes #115

## Checklist:

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -15,6 +15,7 @@ from .util import get_ip_address
 
 # Subscription event types.
 EVENT_TYPE_BINARY_STATE = "BinaryState"
+EVENT_TYPE_INSIGHT_PARAMS = "InsightParams"
 EVENT_TYPE_LONG_PRESS = "LongPress"
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


Spectre5:
Hey - just seeing this and curious why you closed it?  Was it not working reliably or something else?

esev:
It's broken. It didn't properly handle the case where the device is reconnected.
https://github.com/pavoni/pywemo/blob/820f29cba1a1939da62fce5ff028e3d3eadf5911/pywemo/subscribe.py#L300

I also wanted to split these changes up a bit.

I've got it working locally now and will send the first PR shortly.

esev:
Here's the new PR that fixes the issue with this one.
https://github.com/pavoni/pywemo/compare/master...esev:fix_sched

It depends on #210, so I'll send it after that one is merged.

esev:
I guess I can't reopen this PR.  Moved to #227

PR #210, opened 2021-01-26
SubscriptionRegistry bug fixes (timeout/exception handling/headers/sched)

esev:
## Description:

A few bug fixes related to the SubscriptionRegistry

- 12d43d0 (Add a 10 second timeout for resubscribe requests): Connectivity issues can causes http requests to get stuck. The default for the `requests` library is to wait indefinitely for a request to complete. This can cause the entire `Wemo Events Thread` to become stuck. Fix this by adding a small timeout on every request. A 10 second timeout was chosen as that value appears elsewhere in pyWeMo
- d54d85a (Catch RequestException & ActionException when reconnecting for discovery): An uncaught `requests` exception that occurs on the ` Wemo Events Thread` during the call to [reconnect_with_device](https://github.com/pavoni/pywemo/blob/820f29cba1a1939da62fce5ff028e3d3eadf5911/pywemo/subscribe.py#L300) can cause the thread to exit and further subscriptions to stop working.
- ae11b00 (Fix timeout header): `TIMEOUT: 300` isn't a valid header for UPnP subscriptions. See [page 64](http://www.upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.0-20080424.pdf). Use `TIMEOUT: Second-300`. Also default to using 300 if the response doesn't contain a TIMEOUT header.
- 5d29011 (Remove pending events on unregister): When a device is unregistered from the SubscriptionRegistry its pending events are not cancelled. This continues to keep the subscription alive even after the device has been unregistered. It also prevents the `stop` method from being able to terminate the thread. Fix this by removing the pending event on unregister.

This PR also adds tests for the SubscriptionRegistry to validate these changes.

I've been testing against my devices with this script:

```python
"""Test script."""
import logging
import time

import pywemo

WEMO_IP = ''  # <-- fill this in

def _cb(_device, _type, _params):
    logging.info('%s %s %s', _device.name, _type, _params)

sub = pywemo.SubscriptionRegistry()
sub.start()

device = pywemo.discovery.device_from_description(
    pywemo.discovery.setup_url_for_address(WEMO_IP, 0), None
)

sub.register(device)
sub.on(device, None, _cb)

try:
    logging.info('Ready. Press CTRL-c to exit')
    time.sleep(999999)
finally:
    logging.info('Shutting down')
    sub.unregister(device)
    sub.stop()
```

## Checklist:

### DIFF
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -7,6 +7,7 @@ import requests
 
 from . import ssdp
 from .ouimeaux_device import UnsupportedDevice, probe_wemo
+from .ouimeaux_device.api.service import ActionException
 from .ouimeaux_device.api.xsd import device as deviceParser
 from .ouimeaux_device.bridge import Bridge
 from .ouimeaux_device.coffeemaker import CoffeeMaker
@@ -45,12 +46,17 @@ def discover_devices(
             {'manufacturer': 'Belkin International Inc.'}
         ):
             mac = entry.description.get('device').get('macAddress')
-            device = device_from_description(
-                description_url=entry.location,
-                mac=mac,
-                rediscovery_enabled=rediscovery_enabled,
-            )
-
+            try:
+                device = device_from_description(
+                    description_url=entry.location,
+                    mac=mac,
+                    rediscovery_enabled=rediscovery_enabled,
+                )
+            except (requests.RequestException, ActionException) as exc:
+                LOG.warning(
+                    'Could not connect to device %s (%s)', entry.location, exc
+                )
+                device = None
             if device is not None:
                 wemos.append(device)
 
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -188,11 +188,16 @@ class Device:
         port = probe_device(self)
 
         if port is None:
-            LOG.error('Unable to re-probe wemo at %s', self.host)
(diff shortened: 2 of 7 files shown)


esev:
Moving this PR to draft. I'm not sure 9961eb9 is the right fix. There are several calls to reconnect_with_device in pyWeMo as well as in HomeAssistant. I think it's better to make sure reconnect_with_device doesn't raise the RequestException in the first place. That way there is no need to wrap every call to reconnect_with_device with exception handling.

I'm testing a change that does just this in my HomeAssistant instance now. Will move this out of draft tomorrow - assuming the testing goes okay.

esev:
d54d85a is the better fix. I've seen no uncaught exception errors in my logs. With 9961eb9 I was seeing about 4/hour from the other callers of reconnect_with_device.

Moving back to 'ready for review'.

esev:
Rebased with the master branch.

pavoni:
This looks good to me. A significant improvement!

I tried it in my test version of HA - and it seemed to work fine.

I think you've already soak tested it over a few days?

esev:
Thanks for taking a look!

Yes, this PR + #213 & #209 together have been running very reliably for about 5 days. No freezes/hangs of the subscription logic have happened with these updates. There have also been no more exceptions in the HA update polling logic (fixed by d54d85a).

I'll merge this now and reopen #209.

Issue #114, opened 2018-12-20
Rediscovery of devices when IP or port changes is not always consistent


sqldiablo:
WeMo devices periodically update their port number and IP address based on DHCP or other network setup and conditions. Pywemo attempts to rediscover devices when this happens, and will often find the device again and allow it to keep working. However, 1) rediscovery doesn't always work, 2) rediscovery almost always fails the 2nd time around for a given device when it changes its IP or port yet again, and 3) subscriptions don't seem to follow the device during a rediscovery and instead stop working altogether or report warnings for invalid subscription updates for devices we aren't subscribed too (because the device seems "new" to pywemo, since it has a different IP/port than before).

sqldiablo:
I'm actively working on a different approach to handling rediscovery that will just update the existing device object instance, instead of overwriting the entire object instance with another. Once I get that working, I'll take a look at the subscriptions part of the issue.

sqldiablo:
Pywemo also reports that it has found malformed XML during SSDP discovery due to other devices on the network responding to our discovery requests. I'll be removing the error, as it's not relevant and isn't really an error condition.

elmigbot:
Hello, not sure if this is related, but:

When I unplug one wemo plug (mini), it gets randomly identified as any of my other wemo products and continues to fail re-discovery. After re-plugging, the only way to find the correct switch again, is to restart HA.

**Unplugged for a minute, then replugged:**

WARNING (SyncWorker_5) [pywemo.ouimeaux_device.api.service] Error communicating with Guest Fan, retry 0
ERROR (SyncWorker_5) [pywemo.ouimeaux_device] Unable to re-probe wemo at 192.168.1.10
INFO (SyncWorker_5) [pywemo.ouimeaux_device] Trying to reconnect with Guest Fan
INFO (SyncWorker_5) [pywemo.ouimeaux_device] Found Guest Fan again, updating local values [Updated to my Dining Room Dimmer @ 192.168.1.11]
WARNING (SyncWorker_5) [pywemo.ouimeaux_device.api.service] Error communicating with Dining Room, retry 1
INFO (SyncWorker_5) [pywemo.ouimeaux_device] Reconnected to wemo at 192.168.1.11 on port 49153
WARNING (SyncWorker_5) [pywemo.ouimeaux_device.api.service] Error communicating with Dining Room, retry 2
INFO (SyncWorker_5) [pywemo.ouimeaux_device] Reconnected to wemo at 192.168.1.11 on port 49153
[...]

PR #211, opened 2021-01-26
Reset UPnP description cache each scan

esev:
## Description:

Connectivity issues can cause a SSDP scan to fail to read a device's /setup.xml file. Because of the way the /setup.xml is cached, this can make it so that a device can never be discovered.
https://github.com/pavoni/pywemo/blob/820f29cba1a1939da62fce5ff028e3d3eadf5911/pywemo/ssdp.py#L186-L190

Multiple SSDP M-SEARCH messages are sent as part of the discovery scan. I believe the original intent of the `DESCRIPTION_CACHE` was to avoid polling the same device more than once during the same discovery scan call. That behavior is preserved here. The new behavior is to use a fresh cache for each scan.

## Checklist:

### DIFF
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -223,6 +223,11 @@ class UPNPEntry:
             }
         )
 
+    @classmethod
+    def reset_cache(cls):
+        """Clear the internal cache of device descriptions."""
+        cls.DESCRIPTION_CACHE = {'_NO_LOCATION': {}}
+
     def __eq__(self, other):
         """Equality operator."""
         return (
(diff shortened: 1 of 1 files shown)


Spectre5:
Good point here.  If the element tree ends up as `None`, then with the current code that url would permanently be an empty dictionary.

As an aside, while looking at this, isn't there a few other potential code clean ups that could happen here.  For example, `if self.description is None:` can never be True, since the description returns an empty dictionary for unknown/bad URLs.  Not sure where the best place to have conversations about non-PR related code is as neither an issue nor a PR seems right for just the discussions.  Maybe we should enable GitHub Discussions for this type of thing @pavoni.

pavoni:
@Spectre5 Happy to enable discussions - although in this case it feels like we could just open a "Discovery could be improved" issue!

esev:
I wonder if this fixes #114.

PR #212, opened 2021-01-26
Ignore the virtual device when scanning

esev:
## Description:

The virtual device is not a fully implemented device. Filter it out from the SSDP discovery scan as there is no pyWeMo device class that can be associated with it.

## Checklist:


Spectre5 (approved):
I think that the `discover_devices` function won't have any issue as the manufacturer is `pywemo`.  But `scan` could be used outside of that function and then it makes sense to drop it since we know it is fake.

PR #213, opened 2021-01-27
Add timeout and threading for serving http requests

esev:
## Description:

The current `RequestHandler` for receiving notifications from devices is single-threaded and has no socket timeouts. If a single device goes offline while it is connected to the RequestHandler, the handler will block forever and no longer respond to any other devices.

Fix this in two ways:

- Add a 10 second timeout for all sockets. This ensures the connection will be dropped if the device/client gets disconnected while in the middle of a request.
- Use a threaded HTTP server to allow multiple connections at the same time.

## Checklist:

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -5,7 +5,7 @@ import sched
 import socket
 import threading
 import time
-from http.server import BaseHTTPRequestHandler, HTTPServer
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 
 import requests
 from lxml import etree as et
@@ -66,7 +66,7 @@ def _start_server():
     for i in range(0, 128):
         port = 8989 + i
         try:
-            return HTTPServer(('', port), RequestHandler)
+            return ThreadingHTTPServer(('', port), RequestHandler)
         except (OSError, socket.error):
             continue
     return None
@@ -117,6 +117,9 @@ class RequestHandler(BaseHTTPRequestHandler):
       from the pywemo SubscriptionRegistry. The event type for a long press
       action is EVENT_TYPE_LONG_PRESS.
     """
+
+    # Do not wait for more than 10 seconds for any request to complete.
+    timeout = 10
 
     def do_NOTIFY(self):  # pylint: disable=invalid-name
         """Handle subscription responses received from devices."""


Spectre5 (approved):

Spectre5:
Makes sense, you're waiting for other PR to get merged first?

esev:
> Makes sense, you're waiting for other PR to get merged first?

I wanted to give it a bit more time to soak in my HomeAssistant instance. Just looking out for thread exhaustion or some other mistake. So far so good though, I haven't seen the handler block with this PR and I don't see any increase in the thread counts. It's making the notifications/long-press work a lot more reliably though!

I'll move this out of draft tomorrow morning - assuming all is still running well.

Spectre5:
Thread exhaustion was my 1 concern too, I know that overall HA uses a lot of threads!

Before this did the long press sometimes not work then?  I could see this improving the response time for it with the threading, but overall I wouldn't have expected this timeout to come into play frequently.

esev:
I have a few devices that are right on the edge of getting a good Wifi connection. These devices drop out a lot and reset themselves. Every once in a while (about once/week) I notice that all the notifications (BinaryState/InsightParams/LongPress) stop working - even for the devices with a solid wifi connection. I could see that pyWeMo was still trying to maintain the subscriptions (it was still sending SUBSCRIBE/UNSUBSCRIBE), but no notifications actually worked. That's when I looked at the network with tcpdump and found that the handler had stopped responding. With all the requests for /setup.xml, the subscriptions, and the notifications, it looks like all the right conditions happen where a device loses connection while it is connected to the handler.

I've been taking a deeper dive into each of the causes when this happens. That's why the recent flurry of PRs.

esev:
Come to think of it, it was probably #209 that was making these issues happen more often for me. One of the devices with a weak Wifi connection is an Insight device. InsightParams notifications happen about once/second so it has likely made these issues more pronounced (and a bit easier to reproduce).

I have all of these PRs (including the multiple subscriptions) running in my HomeAssistant instance right now. I haven't seen any more of the issues where the handler hangs since adding this PR. Hopefully it stays that way overnight :)

esev:
Everything looks good in my HomeAssistant. All the notifications are working after 16 hours and I see no significant increase to the number of threads.

PR #217, opened 2021-01-29
Consider the request body in VCR tests

esev:
## Description:

Consider the request ['method', 'scheme', 'host', 'port', 'path', 'query', 'body'] in VCR tests.

The previous default was ['method', 'scheme', 'host', 'port', 'path', 'query']

## Checklist:

### DIFF
--- a/tests/vcr/tests.ouimeaux_device.test_bridge/test_light_turn_off.yaml
+++ b/tests/vcr/tests.ouimeaux_device.test_bridge/test_light_turn_off.yaml
@@ -63,7 +63,7 @@ interactions:
 
       <u:GetEndDevicesWithStatus xmlns:u="urn:Belkin:service:bridge:1">
 
-      <DevUDN>uuid:Bridge-1_0-231438B01000D9</DevUDN>
+      <DevUDN>uuid:Bridge-1_0-SERIALNUMBER</DevUDN>
 
       <ReqListType>PAIRED_LIST</ReqListType>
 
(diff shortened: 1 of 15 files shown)


tests/vcr/tests.ouimeaux_device.test_bridge/test_light_turn_off.yaml
```diff
@@ -123,7 +123,7 @@ interactions:
 
       <u:SetDeviceStatus xmlns:u="urn:Belkin:service:bridge:1">
 
-      <DeviceStatusList>&lt;?xml version=&#x27;1.0&#x27; encoding=&#x27;utf-8&#x27;?&gt;
+      <DeviceStatusList>&lt;?xml version=&#x27;1.0&#x27; encoding=&#x27;UTF-8&#x27;?&gt;
```
Spectre5, line 126: Just curious, is this an actual change from re-running vcr?  Why was it lowercase to start with?

Spectre5:
Without looking at it any closer at the moment, is there any reason to not include all of the defaults to match on plus the body?  I'm not advocating either way here, mostly wondering out loud

tests/vcr/tests.ouimeaux_device.test_bridge/test_light_turn_off.yaml
esev: Yes, the tests were failing without this change. I'm not sure if this is related to #202 or something else.

esev:
> Without looking at it any closer at the moment, is there any reason to not include all of the defaults to match on plus the body?

hehe. I added the defaults to the PR description because I was wondering out loud too. I thought it might make it easier to record if we ignore the host/port. But maybe it is good if we don't so we can catch mistakes. I'll change it to use the defaults + body.

tests/vcr/tests.ouimeaux_device.test_bridge/test_light_turn_off.yaml
Spectre5: ah, ya, that'd make sense, since the code actually does still have lowercase `uft-8` in it, so lxml must change it to uppercase.  Maybe  for consistency we should change the TEMPLATE to be uppercase too in this PR

esev: I did a git revert e375e12f4901461b02ceb08feb5c02a766c560ee and ran the test again. With #202 reverted I have to use lower-case 'utf'. So it seems that PR is where the change happened.

It doesn't seem to make a practical difference though. My bridge device works either way.

esev: But it sure does validate your reasoning for why we want to check the request body. It would have been nice to know about this change earlier.

Spectre5:
I think we should change that TEMPLATE `uft-8` to `UFT-8` to make it consistent with what `lxml` is doing then.  Assuming that doesn't cause an issue then I think this is good to merge.

esev:
Sorry, I missed that in your original comment. Changed it so the case matches.

Spectre5:
I probably edited that in after you read it :)

PR #218, opened 2021-01-29
add coverage

Spectre5:
## Description:

With the addition of new tests and plans for more, I'd like to include coverage.  @pavoni, you might consider signing up for coveralls too (and could add a badge to the README), but it isn't that important either.

Needed to bump pytest to over 6.0 so it can read from pyproject.toml (putting outputs into .cache/pytest now), and obviously added coverage dependency and a few options for it to pyproject.toml too.

## Checklist:

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -97,6 +97,20 @@ optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [[package]]
+name = "coverage"
+version = "5.4"
+description = "Code coverage measurement for Python"
+category = "dev"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4"
+
+[package.dependencies]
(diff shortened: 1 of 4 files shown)


esev (approved):
Nice idea

Spectre5:
Added changing one of the regex in `conftest.py` to a raw string to quiet/fix this pytest warning:

```
tests/conftest.py:22
  /home/scott/Coding/github/pywemo/tests/conftest.py:22: DeprecationWarning: invalid escape sequence \d
    b'uuid:([A-Z][a-z]+-\d_\d)-[A-Za-z0-9]+',
```

pavoni:
I've signed up for coverage.io - and added the token to the github secrets `CODECOV_TOKEN`.

I started to look at how to submit - and then ran out of time. But if you already know how to do this - please go ahead!

Spectre5:
You mean coveralls.io?  If so, you can just name the secret `COVERALLS_REPO_TOKEN` and then we just add coveralls as another dev dependency, then call `coveralls` after `coverage`.  I'll take a closer look later as I also don't have time right this minute!

pavoni:
Now added as `COVERALLS_REPO_TOKEN` (just while it was fresh in my mind - no rush!).

PR #219, opened 2021-01-29
Verify device IDs when probing

esev:
## Description:

Verify device IDs when probing. This is to avoid an [edge case](https://github.com/pavoni/pywemo/issues/114#issuecomment-450521329) with device IPs changing.

**Related issue (if applicable):** #114

## Checklist:


pavoni:
LGTM

Issue #67, opened 2017-02-18
Change subscriptions to update cached params 


pavoni:
When using subscriptions it would be good to use the data in the subscription callbacks - rather than force the client to refetch the data.

Particularly useful now insight params are updated every second for an active device.

@jumpkick's recent PR points the direction in adding the insight subscription.

I suspect this means we'll want to subscribe to multiple events types for devices that support then (eg insight switches have `basic` and `insight` events) - but only once we stop calling back to the device for every update.

0.4.13 traces what the callback type is - and params that were passed - so should make reverse engineering easier!.

@stu-gott's code for the coffeemaker has the basics which we could adapt / extend for other devices.

@jaharkes Suspect the bridge will be the most complex!

pavoni:
I suspect we should now close this @esev do you have a view

esev:
I think it can be closed too. I think the bridge might be the only device that doesn't do this at the moment. But the bridge is also not setup for a subscription in home assistant either.

PR #194, opened 2021-01-01
Add long press support to LightSwitch and Dimmer devices

esev:
## Description:

This PR exposes the functionality to configure long press actions for the LightSwitch and Dimmer devices.

When combined with #192 & #193 it can be tested with the following script:

```
import logging
import time

import pywemo

WEMO_IP = ''  # Fill this in

def _cb(_device, _type, _params):
    logging.info(
        '%s %s %s', _device.name, _type, _params
    )

logging.basicConfig(level=logging.DEBUG)
sub = pywemo.SubscriptionRegistry()
sub.start()

responder = pywemo.ssdp.DiscoveryResponder(sub.port)
responder.start()

device = pywemo.discovery.device_from_description(
    pywemo.discovery.setup_url_for_address(WEMO_IP, 0), None
)

device.ensure_long_press_virtual_device()
sub.register(device)
sub.on(device, pywemo.subscribe.EVENT_TYPE_LONG_PRESS, _cb)

try:
    time.sleep(999999)
finally:
    responder.stop()
    sub.stop()
```

**Related issue (if applicable):** fixes #61 

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] There is no commented out code in this PR.

PR #220, opened 2021-01-29
Update the dimmer attributes due to a subscription update event

esev:
## Description:

Update the _state & _brightness cached attributes based on a subscription update.

**Related issue (if applicable):** #67

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -43,7 +43,7 @@ class Dimmer(Switch, LongPressMixin):
 
     def subscription_update(self, _type, _param):
         """Update the dimmer attributes due to a subscription update event."""
-        if _type == "Brightness":
+        if _type == "Brightness" and self._state:
             self._brightness = int(_param)
             return True
         return super().subscription_update(_type, _param)
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -42,8 +42,11 @@ class Dimmer(Switch, LongPressMixin):
         self._brightness = int(brightness)
 
     def subscription_update(self, _type, _param):
-        """Disable subscription updates."""
-        return False
+        """Update the dimmer attributes due to a subscription update event."""
+        if _type == "Brightness":
+            self._brightness = int(_param)
+            return True
+        return super().subscription_update(_type, _param)
 
     def __repr__(self):
         """Return a string representation of the device."""
--- a/tests/ouimeaux_device/test_dimmer.py
+++ b/tests/ouimeaux_device/test_dimmer.py
@@ -22,11 +22,14 @@ class Base:
         assert dimmer.get_state(force_update=True) == 0
 
     def test_subscription_update_brightness(self, dimmer):
-        assert dimmer.subscription_update('Brightness', '23') == True
-        assert dimmer.get_brightness() == 23
+        # Does not update when the light is off.
+        assert dimmer.subscription_update('Brightness', '23') == False
 
         assert dimmer.subscription_update('BinaryState', '1') == True
         assert dimmer.get_state() == 1
+
+        assert dimmer.subscription_update('Brightness', '52') == True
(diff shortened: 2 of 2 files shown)


pavoni:
Code looks good. Can't test it myself - but I can see you have!

esev:
I've tested with both fd001d1 alone & with 6f802a8. HomeAssistant works fine either way.

I don't understand why the `set_brightness` method has this check. Why is it important that the BinaryState change when the Brightness changes?
https://github.com/pavoni/pywemo/blob/8a1e77ba60024defe4320dabd5add33025dd6041/pywemo/ouimeaux_device/dimmer.py#L26-L38

Given the above check, it seems like it is important that the _brightness value be 0 when the device was off. I added the extra `and _state` check because I thought maybe some (non-HA) client requires this.

Personally I'd like to be able to set the Brightness independently from the BinaryState. It would be nice to be able to set some of my lights to a dimmer level in the evening without needing to turn them on. That way I can preset the brightness level for when someone presses the physical button on the device. 

Example: I'd like the bedroom light to come on dim after 10pm when someone presses the physical button. I don't want to have to turn the light on to make that happen. If someone was already asleep in there that would be annoying :)

@pavoni do you have more background on this? Do you think the extra `and _state` check (6f802a8) is required by any client?

pavoni:
It may be some sort of HA compatibility. 

I don't have a Wemo dimmer - but I have some zwave/vera dimmers.

They don't have the logic above - but they behave a bit as if they do.

Setting a brightness level of > 0 turns on the switch (although setting a brightness level of 0 doesn't turn off the switch).

I'm open to look at changing this - but best to do it in another PR (would be good to check against a few other HA devices - or move the logic if needed into the HA wrapper).

esev:
> would be good to check against a few other HA devices - or move the logic if needed into the HA wrapper

+1 - I'm not sure HA supports setting the brightness without turning on the light either. It is probably not a common use-case.

> but best to do it in another PR 

Definitely. I might end up just doing something in a custom component. Not sure if this would be generally useful.

Spectre5:
I guess it isn't a common use case.  When I use my dimmer manually, I often adjust the brightness up or down with my finger before turning it on.  I support HA thinks that if you're going to press the button anyway, then you can also manually adjust it first if that's your desire, since turning it on via automation with HA allows you to set it directly at the same time.  I think that if it is easy enough to support, there isn't any reason not to though.  I know that HA would have to support it too though, not just pywemo.

esev:
It's all good. I was more interested in learning the background on why it is the way it is. Sounds like WeMo dimmers are a bit different in that they let the brightness be controlled separately from the on/off state. It won't be in the best interest of HA to support this generically if it only works on a small number of devices.

It sounds like this will only be useful to me. I added a [custom patch in my local instance](https://github.com/pavoni/pywemo/pull/194#issuecomment-757353570) to light.py that exposes a [wemo.set_dimmer_brightness](https://github.com/esev/home-assistant-core/commit/9a51421c5442b27890e4731e567dcc67553bcfd0) service. The service calls `self.wemo.basicevent.SetBinaryState` directly.

Back to this PR & #67  :)  @Spectre5 IIRC you have a dimmer. Does this PR look okay to be merged from your perspective?

pywemo/ouimeaux_device/dimmer.py
```diff
@@ -42,8 +42,11 @@ def set_brightness(self, brightness):
         self._brightness = int(brightness)
 
     def subscription_update(self, _type, _param):
-        """Disable subscription updates."""
-        return False
+        """Update the dimmer attributes due to a subscription update event."""
+        if _type == "Brightness" and self._state:
```
Spectre5, line 46: My only question is whether this should be `Brightness` or `brightness`.  From the dimmer YAML file, I see `<relatedStateVariable>brightness</relatedStateVariable>`

Spectre5: ah, never, the wemo does send `Brightness`

Spectre5:
Looks good to me

PR #221, opened 2021-01-30
setup pydocstyle

Spectre5:
## Description:
I'm not sure if we actually want to use pydocstyle since I'm not sure that it really provides anything extra over flake8 and pylint.  However, since I saw it in the dev dependencies but not used in the project, I threw in a quick configuration and a couple of edits to quiet it.  The few edits that I made would have been caught by flake8 and/or pylint too, except that we don't run them on the tests directory.  So if I also excluded the tests directory, then pydocstyle wouldn't have caught anything.  It does, by default, exclude files that start with `test_`, but there are files in the tests directory that don't start with test.

If we don't want to include it, then I'm fine to cancel this PR and then we can just remove that dev dependency instead.  Any preferences @pavoni and @esev regarding keeping or removing it?

## Checklist:


tests/conftest.py
```diff
@@ -1,3 +1,4 @@
+"""Configuration file for VCR."""
```
esev: """Shared pytest fixtures.""" ?

tests/ouimeaux_device/api/unit/long_press_helpers.py
```diff
@@ -1,3 +1,5 @@
+"""Helper(s) for longer press testing on supported devices."""
```
esev: Nit: "long press feature"

esev (commented):
Is it possible to run flake8/pylint for the tests directory, or would that require a lot of work?

I'm leaning toward removing the dependency if flake8/pylint have the same level of coverage for pydocs and they can be enabled for the tests directory.

Spectre5:
They can easily be adjusted to include the tests directory.  Pylint will make a ton of complaints due to how tests are normally written, but we could even use just flake8 for those, or nothing like now.

esev:
Wow, yeah, pylint has a lot of complaints about the pytest fixtures used in the tests directory!  flake8 looks like it has good checks for the pydoc strings.

I'm all for trying to keep all the code consistent. So I think its worthwhile to apply the code checks to the tests directory. I think I'd lean toward using flake8 rather than pydocstyle given the large overlap in the results. But you likely have more experience with these tools than I do. Are there advantages to pydocstyle over flake8?

Spectre5:
This is actually my first time using pydocstyle, but I have used flake8 and pylint quite a bit.  It appears to me that pydocstyle doesn't add anything extra over flake8, so I think I'd lean towards removing it as well.  But part of the reason I added this PR was to see what you and pavoni thought, in case you'd had experiences with pydocstyle.

To include flake8 in the tests directory, it would be very easily -- simple remove the line with `tests` in it under the exclude section in `setup.cfg`.  For pylint, we'd change `pylint pywemo` to `pylint pywemo tests`to include it, but using pylint on the tests would require a lot of disables.

I think my preference would probably be to remove pydocstyle, leave pylint as-is, and then do add flake8 to the tests directories.

Spectre5:
I should also note, that flake8 allows per-file ignores, so you could ignore certain messages in the tests that you don't ignore in the main code, via an option like below in the `setup.cfg` file.

```
[flake8]
per-file-ignores =
    tests/*: D101
```

I think some people prefer no docstrings for tests as the name is often clear enough.  So we could, for example, disable those messages, or line length messages, etc for only the tests directory.

tests/conftest.py
Spectre5: Ya, much better ha

tests/ouimeaux_device/api/unit/long_press_helpers.py
Spectre5: Happy to include nit picks, I like things to be as "correct as possible", when possible

pavoni:
We could move the fixtures into a subdirectory, and exclude it as needed.

I think adding flake8 and pylint to tests is a great idea. 

Don’t have strong feelings about pydocstyle.

The build scripts were contributed, and were clearly used in another Repo. We left the unused infrastructure- but just turned it off.

Now we have a more robust setup - very happy to make a call and remove pydocstyle.

Spectre5:
I think we're all on the same page.  I'll setup a new PR to remove pydocstyle and include flake8 and/or pylint in the test directory.

pavoni:
Thanks!

Spectre5:
Funny, I missed that `flake8-docstring` *is* in the dev requirements though!  That is a plugin for `flake8` to include docstring checking with `pydocstyle`, so we've been using it without realizing it.  This explains why using `pydocstyle` didn't find anything new on files that are already checked by `flake8`!

`pydocstyle` is obviously a dependency of that plugin, so when I make the new PR I'll still remove it, but leave `flake8-docstring`.

Spectre5:
I'm starting on the PR to add flake8 (and *mabye* pylint) to the tests.  But what are your guys' thoughts on requiring method docstrings in the tests?  Do we want to require that?  A couple of examples we have right now are:

```python
    def test_turn_off(self, dimmer):
        """Turn off the dimmer."""
```

```python
    def test_setup_ap_not_found(self, device):
        """Test device setup (APNotFound)."""
```

```python
    def test_reconnect_with_device_by_probing_Timeout(self, get, device):
        # no docstring example
```

The docstring doesn't hurt, but these two examples with it really don't add anything either.  And the one without it is clear enough too.  I feel like a fair number of other test code I've seen tend to leave them off to reduce some verbosity, I assume.  And if disabled, it doesn't mean you can't add them where they are useful.  Although it sounds like I'm arguing against them, I'm actually not (I've added them on all the tests I've written!), but I just want to try to follow common best practices.  So I'm actually pretty indifferent.

Paging @esev and @pavoni, just to make sure you guys notice this

esev:
I think I saw the PR go in for this already. :) My preference is the same as yours: don't require the doc strings on individual test cases but allow them when it helps to understand the test case.

Spectre5:
Yup, that's what I ended up doing!  So add it if the method name isn't clear enough or if the test is more complicated, but don't generally require them.

PR #223, opened 2021-01-30
add coveralls

Spectre5:
## Description:
First attempt at adding coveralls to the github flow.

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,16 +13,17 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: [3.7, 3.9]
-
     steps:
-      - uses: actions/checkout@v1
+      - name: Checkout code
+        uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


Spectre5:
@pavoni, I'll probably let you take it from here :)

Since I'm in these files, I went ahead and updated the github actions for checkout and setup-python both from v1 to v2.

I've never added coveralls to a flow before, so I'm not sure if this will work.  But if this is merged and messes the workflows/github actions, then you'll need to push to master and ignore the CI.

pavoni:
I've reverted this PR for the moment since there were some issues:-

https://github.com/pavoni/pywemo/runs/1797141260?check_suite_focus=true

Happy to have another go at some point

Spectre5:
I see, at that point in the script, the virtual environment is not still active and thus coveralls is not on the path.  The easiest thing to do is probably to just use the `ci` environment variable and put it in the build script.

PR #226, opened 2021-01-30
Add coveralls, take 2

Spectre5:
## Description:
See #223, I've added coveralls to the build script instead.  It would get run on pull requests as well now, but I don't think that causes an issue.

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,11 +1,9 @@
 name: Build
-
 on:
   pull_request:
   push:
     branches:
     - master
-
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


pavoni:
I'll merge this and then we can check it works.

If it does I think it would be pretty easy to change it to only run when we publish - which is perhaps better.

pavoni:
Fantastic - thanks.

Let me know what you think about just running it on publish.

Spectre5:
Right.  So right now it runs on a PR, on push, and on release (publish).  In the README, you can select to show whichever badge you want from shields.io, so I think it is fine to run coveralls on all of those, and then you can display in the README whichever one you want.  I think showing the coverage from the current state of the master branch makes the most sense though.

pavoni:
That's perfect. And agree the `master` branch makes most sense.

PR #227, opened 2021-01-30
Support multiple subscriptions for the Insight device

### LINKED ISSUE #115: Allow pywemo to subscribe to multiple services

esev:
## Description:

Support multiple subscriptions for the Insight device.

The original subscription logic stored the scheduler Event instances in the `self._events` dictionary, keyed by the device serial number. This only permitted one subscription (one scheduler Event) per device.

Original: `self._events[device.serialnumber] = self._sched.enter(...)`

This PR adds an additional level of dictionary. `self._events` is still keyed by serialnumber, but rather than a single Event instance, it now contains a dict with multiple Event instances. Each individual Event is keyed by the function used to fetch the URL for the subscription.

New: `self._events[device.serialnumber][url_fn] = self._sched.enter(...)`

Two new functions are added, each for a specific subscription URL.

```python
def _basic_event_subscription_url(device: Device) -> str:
    """Return the basic event subscription URL."""
    return device.basicevent.eventSubURL


def _insight_event_subscription_url(device: Device) -> str:
    """Return the insight event subscription URL."""
    return device.insight.eventSubURL
```

These functions are the keys for the second-level dictionary. They are also used to lookup the appropriate device URL to maintain the subscription.

It was important in this PR to not store the URL of the device directly, but rather read it from the device instance each time. This is necessary to support probing/rediscovery of a device when its IP/port changes. PR #209 made the mistake of using the URL as the second-level dictionary key and storing it in the Event arguments. #209 was flawed because the URL/host/port never changed when the device was rediscovered. Using the url_fn avoids that mistake.

**Related issue (if applicable):** fixes #115

## Checklist:

### DIFF
--- a/tests/vcr/tests.test_subscribe/Test_SubscriptionRegistry.test_register_unregister.yaml
+++ b/tests/vcr/tests.test_subscribe/Test_SubscriptionRegistry.test_register_unregister.yaml
@@ -39,4 +39,44 @@ interactions:
     status:
       code: 200
       message: OK
+- request:
+    body: null
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      CALLBACK:
+      - <http://192.168.1.1:8989/insight1>
+      Connection:
+      - keep-alive
+      Content-Length:
+      - '0'
+      NT:
+      - upnp:event
+      TIMEOUT:
+      - Second-300
+      User-Agent:
+      - python-requests/2.25.1
+    method: SUBSCRIBE
+    uri: http://192.168.1.100:49153/upnp/event/insight1
+  response:
+    body:
+      string: ''
+    headers:
+      CONTENT-LENGTH:
+      - '0'
+      DATE:
+      - Tue, 26 Jan 2021 02:16:36 GMT
+      SERVER:
+      - Unspecified, UPnP/1.0, Unspecified
+      SID:
+      - uuid:849c1a56-1dd2-11b2-b5fd-dcf7b6ec9aaa
+      TIMEOUT:
+      - Second-300
+      X-User-Agent:
+      - redsonic
+    status:
+      code: 200
+      message: OK
 version: 1
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -310,12 +310,13 @@ class SubscriptionRegistry:
                 if device.rediscovery_enabled:
                     device.reconnect_with_device()
             with self._event_thread_cond:
-                self._events[device.serialnumber] = self._sched.enter(
-                    SUBSCRIPTION_RETRY,
-                    0,
-                    self._resubscribe,
-                    [device, sid, retry],
-                )
+                if self._events.get(device.serialnumber, None):
+                    self._events[device.serialnumber] = self._sched.enter(
+                        SUBSCRIPTION_RETRY,
+                        0,
+                        self._resubscribe,
+                        [device, sid, retry],
+                    )
 
     def _url_resubscribe(self, device, headers, sid, url):
         request_headers = headers.copy()
@@ -336,9 +337,10 @@ class SubscriptionRegistry:
         )
         sid = response.headers.get('sid', sid)
         with self._event_thread_cond:
-            self._events[device.serialnumber] = self._sched.enter(
-                int(timeout * 0.75), 0, self._resubscribe, [device, sid]
-            )
+            if self._events.get(device.serialnumber, None):
+                self._events[device.serialnumber] = self._sched.enter(
+                    int(timeout * 0.75), 0, self._resubscribe, [device, sid]
+                )
 
     def event(self, device, type_, value):
         """Execute the callback for a received event."""
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -17,6 +17,7 @@ from .util import get_ip_address
 
 # Subscription event types.
 EVENT_TYPE_BINARY_STATE = "BinaryState"
+EVENT_TYPE_INSIGHT_PARAMS = "InsightParams"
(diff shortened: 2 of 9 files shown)


tests/vcr/tests.test_subscribe/Test_SubscriptionRegistry.test_register_unregister.yaml
```diff
@@ -39,4 +39,44 @@ interactions:
     status:
       code: 200
       message: OK
+- request:
```
Spectre5, line 42: Empty body and empty response body, is that right?

esev:
@Spectre5 when you have a chance, could you take a look at the build failure? I think it isn't related to this PR, but I could be mistaken.

```
Running on Github Actions but GITHUB_TOKEN is not set. Add "env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}" to your step config.
  File "/home/runner/work/pywemo/pywemo/.venv/lib/python3.9/site-packages/coveralls/cli.py", line 61, in main
    coverallz = Coveralls(token_required,
  File "/home/runner/work/pywemo/pywemo/.venv/lib/python3.9/site-packages/coveralls/api.py", line 47, in __init__
    self.ensure_token()
  File "/home/runner/work/pywemo/pywemo/.venv/lib/python3.9/site-packages/coveralls/api.py", line 54, in ensure_token
    raise CoverallsException(
coveralls.exception.CoverallsException: Running on Github Actions but GITHUB_TOKEN is not set. Add "env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}" to your step config.
Error: Process completed with exit code 1.
```

tests/vcr/tests.test_subscribe/Test_SubscriptionRegistry.test_register_unregister.yaml
esev: That is correct. These requests use only the headers to pass values.

More details can be found in the [UPnP Device Architecture](http://www.upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.0-20080424.pdf) spec starting around page 64.

Spectre5:
Definitely not related to your PR, I'll add that in a new PR right now.

Spectre5:
ok, rebase on master and try again

esev:
@Spectre5 you likely already saw the new build failure. No rush on this. I'm more than happy to rebase/push as many times as necessary so everything can be worked out.

Spectre5:
I suspect that this PR might need to be closed and re-opened or re-created.  I found in one of the various sites I read that after fixing this 422 error (which I hope I have now in master), that they could only get their current PR's to respond correctly after closing and re-opening or re-creating it.

esev:
> I suspect that this PR might need to be closed and re-opened or re-created.

Nice find! Closing and reopening solved it. Thanks!

Spectre5:
whew :sweat_smile: haha

Spectre5:
I haven't really looked at this closely myself yet, what's the status/verdict on it?

esev:
It is working for me. I'd prefer to have a 2nd look at changes like this before merging, but if you think it's okay I can merge it.

Spectre5:
Ok I can take a look at it for logic, just can't physically test it since I don't have an insight!

pavoni:
I'll give it a try - I had a quick look at the code - which looks good. But I'll also run it a little to double check.

Spectre5:
We should add a test_insight file too, to test the methods in the Insight class.  Could tag it on to this PR or make a new one.

pavoni:
This looks good - and works well in my setup.

Are we all happy to merge?

esev:
It has been working well in my HA too. I haven't observed any issues. I'm good to merge.

It sure catches a lot more state changes :)
Screenshot 2021-02-04 at 9 04 15 AM

pavoni:
Thanks @esev, let's check @Spectre5 is happy to, and then I'll merge and release.

It looks like HA benefits quite a bit - even without changes to the HA wrapper - so once we've merged shall I do an HA version bump. It'll be a simple PR so will likely be merged easily (the rules PR is a lot more complex so might take more time to review)?

pavoni:
And BTW thanks for this - as you know I tried (and failed) to do this myself a few years back!

esev:
> once we've merged shall I do an HA version bump. It'll be a simple PR so will likely be merged easily (the rules PR is a lot more complex so might take more time to review)?

Thank you @pavoni! I agree, version bumps go quickly (https://github.com/home-assistant/core/pull/45501), feature requests go slower (https://github.com/home-assistant/core/pull/45503).

If either of you have time to review https://github.com/home-assistant/core/pull/45503, maybe that will speed it along?

esev:
> once we've merged shall I do an HA version bump.

Actually I misread that as "I shall do". :) I'll take care of it.

Spectre5 (approved):

Spectre5:
looks good to me.

PR #228, opened 2021-01-30
add github token to CI env

Spectre5:
## Description:
Adding GITHUB_TOKEN to the CI environment variables.

## Checklist:


Spectre5:
~@pavoni, I think you need to merge this to bypass the CI?~ hm....didn't fail for me like it did in #227

PR #229, opened 2021-01-30
add service github for coveralls

Spectre5:
## Description:
Another fix for coveralls

## Checklist:


Spectre5:
@pavoni, now this one does fail.  I think this needs to be pushed ignoring CI, but I don't think I can do that?

pavoni:
Looks like it didn't help - but happy to do the same again. Just checked if I could give you admin permission while you do this - but sadly not!

PR #230, opened 2021-01-30
coveralls adjustments

Spectre5:
## Description:
When run with parallel jobs, you need a final "finish" step.  I've made a few adjustments based on this:
https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support

That adjustment probably isn't needed for the publish since I don't think it runs parallel like the pull requests/build script does, but I assume it won't hurt since it will still run the `--finish` part.

Also, right now it just installs the latest coveralls in this final/finish step.  If this works, we can make that a little bit smarter to re-activate the virtual environment and use the same version that is already installed.

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,5 +38,5 @@ jobs:
           COVERALLS_SERVICE_NAME: github-actions
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          pip3 install --upgrade coveralls
+          pip3 install --upgrade coveralls setuptools
           coveralls --finish
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
         env:
           CI: 1
           COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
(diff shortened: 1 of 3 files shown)


Spectre5:
ok, the last few commits look to be an issue with out use of poetry instead of setuptools and the [build-system] section of the prproject.toml file.  There is some good information here:
https://github.com/pypa/setuptools/issues/2353

Spectre5:
I don't really want to add setuptools to the [build-requires] section just for this.  I see that there are a few work-arounds mentioned, so I'll try those out a little later today.

Spectre5:
Actually, it is now running but saying the github token is not valid?  Again, will look more later...

```
Finishing parallel jobs...
Parallel finish failed: Invalid github token
  File "/usr/local/lib/python3.9/site-packages/coveralls/cli.py", line 80, in main
    coverallz.parallel_finish()
  File "/usr/local/lib/python3.9/site-packages/coveralls/api.py", line 283, in parallel_finish
    raise CoverallsException('Parallel finish failed: {}'.format(e))
coveralls.exception.CoverallsException: Parallel finish failed: Invalid github token
```

Spectre5:
I think this last issue is actually due to the code not being committed.  So I think that you can accept this PR and ignore this CI error, and then it should work after it is *in* the repository.

PR #231, opened 2021-01-30
add module docstring

Spectre5:
## Description:
Test to see if CI works...

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,6 @@ jobs:
       - name: Build
         env:
           CI: 1
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
           COVERALLS_SERVICE_NAME: github
           COVERALLS_PARALLEL: true
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -35,7 +34,6 @@ jobs:
     steps:
       - name: Finished
         env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
           COVERALLS_SERVICE_NAME: github
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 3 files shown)


pavoni:
Is this any help?

https://github.com/AndreMiras/coveralls-python-action

Spectre5:
I hadn't seen that, could certainly try it instead.  Funny how the error I get here is different than what @esev is getting in his current PR (and I re-triggered the checks to be sure it was current).  I don't see why they would be different for this issue.  I'll continue to look into it, both what I've done and/or trying that action you linked to, which I had not seen.

Spectre5:
Lots of current information about this issue [here](https://github.com/TheKevJames/coveralls-python/issues/252).  Looks like this last change is working, but is currently frozen on "waiting for status to be reported"

Perhaps you need to "remove the outdated check in the protected branch settings of the repository settings", based on [this page](https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157).  I'm not exactly sure what that means though.

Another related question though, what are these two checks that are "Expected".  I don't see how or where they are defined, as the 3 completed ones with check marks are the tree we've defined in the build.yml file.  ~What creates these ones?~ These [status checks](https://docs.github.com/en/github/administering-a-repository/troubleshooting-required-status-checks) are an option in the github settings.
Screenshot from 2021-01-30 15-58-00

pavoni:
I’ve reset the branch protection rules to the new names.

Spectre5:
Just for my own future understanding, can you clarify exactly what that means?  Which new names - was it me adding the "name" to the jobs or something else?

pavoni:
looks like there is a list of recent actions

9FC23C09-0410-40E2-A0EA-1B495AB3F5E3

Spectre5:
ah ok, so yes it must be from me changing the name to "Build and test" in that job.  I think it was just "build", as a default with no specially defined name, before.  Thanks!

Spectre5:
This "name" line is newly added by me.  So before it just defaulted to "build" from the job identifier.
```
  build:
    name: Build and test
```

pavoni:
Thanks for your hard work implementing this.

PR #233, opened 2021-02-01
remove unused files/options

### LINKED ISSUE #232: [Discussion] Hound
@pavoni, does the hound ci actually do anything?  In the YAML config file, python is disabled.  So I assume it is only being used to check bash files syntax (doesn't support reStructuredText)?
https://github.com/houndci/hound

Spectre5:
## Description:
Remove the hound CI file which is not longer used and also remove the unused option from the `setup.cfg` file for a universal wheel.  The universal wheel is for a pure python library supporting python 2 and 3.  When removing this, I didn't notice anything different and I don't think poetry even reads this option anyway.

```
[bdist_wheel]
universal = 1
```

**Related issue (if applicable):** fixes #232

## Checklist:

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,2 +1,0 @@
-python:
-  enabled: false
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,3 @@
-[bdist_wheel]
-universal = 1
-
 [flake8]
 exclude =
     # common directories or files to exclude


Spectre5:
Looks like hound is still enabled in the repo somewhere but gets no response since this PR removes the hound file.

pavoni:
Thanks - hopefully removed the last trace!

PR #234, opened 2021-02-03
add additional linting & add linting to tests

Spectre5:
## Description:
A handful of changes here:

- Moved pylint configuration to the pyproject.toml file and re-organized the file somewhat.  Now all tools are configured in either pyproject.toml or setup.cfg.
- Added some entries to `generated-members` and `ignored-classes` options of pylint so that I could remove all locally disabled checks for these in pylint (all of the `# pylint: disable=maybe-no-member` comments).
- Add flake8 to the tests directory, but do not require docstrings for functions and methods.  Pylint is still only used on the main code, not on the tests.
- Added the max-complexities option to flake8 and semi-randomly assigned a value of 20 to this.  This results in 3 functions being over this limit, which I added local disables for (could refactor them in the future).
- Added flake8-bugbear as a dev dependency.
- Added some configuration options to mypy to the `setup.cfg`, but mypy is *not* enabled right now in the build script and it would currently fail.  This can be tested by activating the environment and simply running `mypy`.
- Removed `socket.error` in favor of `OSError` (see [here](https://docs.python.org/3/library/socket.html#socket.error))
- Most changes in python files were minor adjustments to satisfy the new linting messages.

## Checklist:


esev (approved):
Looks very nice!

PR #235, opened 2021-02-05
Combine all requests timeouts into a single constant

esev:
## Checklist:

### DIFF
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -7,7 +7,7 @@ import requests
 
 from . import ssdp
 from .ouimeaux_device import UnsupportedDevice, probe_wemo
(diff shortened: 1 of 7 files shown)


Spectre5:
Makes sense, should we also add a timeout to all requests?  For example in `rules_db` there is the line `response = requests.get(rule_db_url)`.  I think there are a few others too

Spectre5:
service.py has 1 more too,  `xml = requests.get(url, timeout=10)`

esev:
Nice catch! Thank you.

esev:
> Makes sense, should we also add a timeout to all requests?

I think so. The default is to wait indefinitely, which may block things. Did you see any other places?

Spectre5:
Just those two

PR #236, opened 2021-02-05
Clients can check if a subscription is active

esev:
## Description:

Provide an `is_subscribed` method to allow clients to confirm a device is registered with the SubscriptionRegistry. It will return True only when it has been confirmed that the device can send state updates to pyWeMo.

I'd like to use this in the HomeAssistant component to skip polling for devices that send events.

## Checklist:


pavoni:
Will take a look, but always assumed that HA’s `should_poll` was tested during device startup, rather than dynamically. So not sure this will help.

May be wrong, of course.

pavoni:
But even if it is static not a bad idea to do this in the library.

I didn’t contribute it - but this is done in another HA library I maintain as a device property.

https://github.com/pavoni/pyvera/blob/7cd487d140285bdfb92be04b2b5569d5444a21d9/pyvera/__init__.py#L786

esev:
Thanks for taking a look!

My concern in HA is for devices that are not on the same subnet as HA. The subscription code might not work for these devices, especially if a NAT is involved, as the CALLBACK header could have the wrong IP. That's why I've changed the code to always UNSUBSCRIBE and create a new subscription. This verifies that the device can connect back to pyWeMo.

It does look like HA's [update task](https://github.com/home-assistant/core/blob/9a570d7c3215d1196a8a80f4148d4c19bbdde8ec/homeassistant/helpers/entity_platform.py#L613) checks `should_poll` each time, and I did find [a component](https://github.com/home-assistant/core/blob/9a570d7c3215d1196a8a80f4148d4c19bbdde8ec/homeassistant/components/imap/sensor.py#L109) that appears to set this dynamically. However that doesn't mean it is the _right_ thing to do. My fallback is to check `is_subscribed` in the wemo component's [async_update](https://github.com/home-assistant/core/blob/9a570d7c3215d1196a8a80f4148d4c19bbdde8ec/homeassistant/components/wemo/entity.py#L48) method and return without polling if the device is subscribed.

pavoni:
That’s very interesting - I’m looking at something in Vera where comms errors trigger an update when they happen - but not when they recover. 

So if it works I’ll use the same trick.

pavoni:
BTW I can confirm you were right - and my guess was wrong. Dynamically changing `should_poll` in HA works perfectly!

pywemo/subscribe.py
```diff
@@ -307,20 +311,16 @@ def _resubscribe(
...
-            headers['SID'] = sid
-        else:
-            host = get_ip_address(host=device.host)
-            headers.update(
-                {
-                    "CALLBACK": '<http://%s:%d/%s>' % (host, self.port, path),
-                    "NT": "upnp:event",
-                }
-            )
+        self.devices[device.host] = device  # If rediscovered with a new IP.
```
pavoni: This looks fine - although I think we turned off this sort of shape shifting!

pywemo/subscribe.py
```diff
@@ -351,24 +351,33 @@ def _url_resubscribe(
         sid: Optional[str],  # pylint: disable=unsubscriptable-object
         url_fn: SubscribeUrlFn,
     ) -> None:
-        request_headers = headers.copy()
         url = url_fn(device)
+        if sid:
```
pavoni: Can you explain a bit more about why you're unsubscribing here?

pywemo/subscribe.py
```diff
@@ -351,24 +351,33 @@ def _url_resubscribe(
...
+            try:
+                requests.request(
+                    method='UNSUBSCRIBE',
+                    url=url,
+                    headers={'SID': sid},
+                    timeout=REQUESTS_TIMEOUT,
+                )
+            except requests.RequestException:
+                pass
         response = requests.request(
```
pavoni: Is there a possible window between unsubscribing and resubscribing - where we may miss an update?

pavoni:
Code looks great as always.

As I understand it - this will always unsubscribe and resubscribe when renewing a subscription. The previous code only did this is there was an error of some kind.

Generally this feels less optimal to me - why do you feel it's needed?

esev:
**Current state**:

The SID (Subscription ID) header is used to identify a specific subscribed client. Currently pyWeMo tries to reuse the same SID when renewing a subscription. If that fails for any reason, pyWeMo is _polite_ and attempts to unsubscribe the SID before trying to create a new subscription. Renewing a subscription then only requires one SUBSCRIBE request if everything is successful. Only on a failure are there additional UNSUBSCRIBE and SUBSCRIBE requests. Failures during renewal are rare.

There is currently no check that the SUBSCRIBE request worked beyond looking for the 412 result code; no confirmation of a 200 result code. There is also no confirmation that the device can actually connect to pyWeMo in order to deliver state update events. The logic does work very reliably though.

**Background**:

WeMo devices treat new subscriptions differently from subscription renewals. I'm using 'new subscription' to mean a SUBSCRIBE request without a SID and using 'renewal' to mean a SUBSCRIBE request with a SID. When a new subscription happens, the Wemo device replies to that subscription with a SID. The WeMo device also immediately delivers a NOTIFY event to pyWemo. When a subscription renewal happens, the WeMo device still responds with the SID, but it does not deliver a NOTIFY event. 

This isn't WeMo specific logic, it is described in the [UPnP spec](http://www.upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.0-20080424.pdf) also.

New Subscription:
> If there are enough resources to maintain the subscription, the publisher should accept it. To accept the subscription, the
publisher assigns a unique identifier for the subscription, assigns a duration for the subscription, **and sends an initial event
message**

Renewal:
>  To accept a renewal, the publisher reassigns a duration for the subscription and must send a response in the same format and
with the same conditions as a response to a request for a new subscription, **except that the initial event message is not sent again**. 

**Goal**: 

For this PR, I want to definitively confirm that 1) the SUBSCRIBE method worked, 2) that the WeMo device is capable of connecting to pyWeMo and delivering notifications and 3) that 1 & 2 remain the case throughout the time-frame from `register` to `unregister`. I'm using the `is_subscribed` method to indicate that all three are true.

**Design**:

I'm using the timestamp of the most recent event notification (`self._last_event_time`) + the timeout value (`self._subscription_timeout`) of the subscription as a check to ensure cases 1, 2 & 3 are all true. With the exception of Insight devices, event notifications are rare and only happen normally when the device is operated. To make this work, I need to ensure there is a event notification within the timeout window. The only way I'm aware of to cause an event notification is to create a new subscription each time.

I'm making a UNSUBSCRIBE request, before creating to a new subscription, to avoid using too many resources on the device. I want to make sure I close the previous subscription before starting a new one.

Technically I could treat the subscription renewal as confirmation that the subscription is still alive. In the case where the renewal succeeds I could update the `self._last_event_time` timestamp directly. But the renewal doesn't cause an actual event notification, so it isn't as strong of a confirmation that the device is still functioning properly and can still reach pyWeMo to deliver event notifications. I think it would also complicate the logic, as we'd need an additional state variable to track whether or not an event notification happened each time a new subscription is created - `self._last_event_time` alone wouldn't be enough to track that state.

**Q&A**:

> As I understand it - this will always unsubscribe and resubscribe when renewing a subscription. The previous code only did this is there was an error of some kind. Generally this feels less optimal to me - why do you feel it's needed?

I agree that it is less optimal for the subscription logic. I'm creating a new subscription each time to get confirmation that the device is still able to connect to pyWeMo to deliver event notifications. And I'm unsubscribing each time to avoid using more than one subscription resource on the WeMo device at a time.

My end goal is to turn off polling in the HA wrapper, but only in situations where it is safe to do so. I've seen HA users with devices on different subnets, which may or may not work with the subscription logic. I feel that this will be a big change, and I want to do my best to not break existing users. To answer your question; the trade-off I see is to add an extra unsubscribe request here every 225 seconds vs making 1-2 requests to update the device state every 10 seconds. It ends up being a net decrease in overall requests to the device. I see that as being more optimal.

The update polling is resulting in some instability that I'm hoping to avoid (https://github.com/home-assistant/core/issues/45904) rather than fix :). I see avoiding polling as an overall win in terms of efficiency too. But if device push fails, I want HA to fallback to using polling. I want to detect if the device push is failing to avoid getting into a situation where no updating happens. And my feeling is that the strongest way to confirm that the device push / subscription logic is working is to get a confirmation each time that pyWeMo can still receive event notifications.

> Is there a possible window between unsubscribing and resubscribing - where we may miss an update?

I believe this is not the case. When a new subscription is created a new event notification also happens. That event notification will contain the current device state. So we shouldn't be missing out on any changes.

> Dynamically changing should_poll in HA works perfectly!

Awesome! Thanks for testing this!

**Additional thoughts**:

I think we can preserve the existing behavior of using subscription renewals if that is desirable. I discussed above that we could treat the subscription renewal as-if it resets the `self._last_event_time` timestamp. But we can also keep the current implementation as the default and provide a boolean (ex`monitor_subscription=False`) that can be set to `True` in the `register` method to always use a new subscription each time. These options do involve more code complexity though.

pywemo/subscribe.py
esev: I added the comment to make this stand out. I didn't want to hide it. It worked! :)

As long as discovery is involved in reconnecting we'll still have some outlier cases where the IP changes and pyWeMo picks up those changes. I felt this change was in line with the 80:20 law (or in this case 99.9:0.1 law) :) It doesn't really impact the 99.9% of cases, and closes one small gap in the 0.1% case.

pywemo/subscribe.py
esev: I added some more details in a [comment below](https://github.com/pavoni/pywemo/pull/236#issuecomment-774325545). And added a link to the comment in the code comments.

pywemo/subscribe.py
esev: I _think_ this is okay and we won't miss an update. I added some [details below](https://github.com/pavoni/pywemo/pull/236#issuecomment-774325545)

pavoni:
Thanks for taking the time to explain the design.

So in effect you’re re-subscribing to force an initial state subscription update.

My experience was that the subscription update mechanism was pretty reliable - and at one point removed the HA refresh, but at the time this didn’t handle the insight parameter updates, so I added it back.

So while I’m not convinced the extra complexity over the regular renewal (which does receive a url success response) is needed - it seems ok as long as it doesn’t cause any other issues.

However I remain concerned about the subscription replacement vs renewal. It makes this a larger, riskier change than is strictly necessary. The current code has been reliable across a range of firmware versions for a long time. 

So I’d rather stick with renewing subscriptions - and if you feel a periodic state refresh is really required to check connection over and above the periodic subscription renewal - do this via a periodic refresh, perhaps less frequently, via HA.

I’d have no issue tracking subscription renewal connection issues and using these to turn on or off HA refreshes dynamically.

Or perhaps adding something in HA allowing refreshes to be turned on if subscription turns out not to work in a particular setup.

esev:
Sounds good. I appreciate your feedback! It makes sense that always unsubscribing and creating new subscriptions is a risky change. I've removed that portion of the code. I've kept it so `is_subscribed` still checks that the event notification happened, but verifying it once when the subscription is first started will likely be enough.

> and at one point removed the HA refresh, but at the time this didn’t handle the insight parameter updates

Your statement made me realize that I wasn't handling multiple subscriptions properly in this PR. For devices with multiple subscriptions the original code in this PR only tracked one subscription. If one failed and the other kept working that would not be detected.

Rather than maintaining more per-subscription state variables in the SubscriptionRegistry, I've factored the subscription code out into it's own `Subscription` class and moved the per-subscription logic and state there. That makes it much easier to manage multiple subscriptions. It also made it a lot easier to improve test coverage of edge cases.

I also fixed a few minor bugs:

- If status_code 412 was always returned from SUBSCRIBE pyWeMo never stopped calling SUBSCRIBE/UNSUBSCRIBE in a tight (recursive) loop. I refactored this a bit so that only one sequence of SUBSCRIBE/UNSUBSCRIBE/SUBSCRIBE can happen each minute.
-  If any error status_code (400-599) other than 412 was returned from SUBSCRIBE this was silently ignored and default parameters were used for `SID` and `TIMEOUT`. This is now checked with `response.raise_for_status()`

pavoni:
This looks good to me. 

Can you confirm when you're happy - then I'll try it in HA to double check - and merge.

esev:
Thank you.

I'm going to work on integrating this into HA too so I can test turning off the polling end-to-end. That way we know it'll work with the next pyWeMo release.

If I make any more changes in this PR, I'll do them as separate commits so the changes are easier to review. I don't anticipate anything major will change; mostly just cleanups/nits.

I'll send another update once I have the new functionality in HA working and tested. Then you can test too and we can merge this PR.

esev:
Small update. It appears that not all WeMo devices are following the UPnP spec. Some do not send the initial event notification when first subscribed. That appears to be the case for the [WSP080](https://www.belkin.com/us/p/P-WSP080/) & [WSP090](https://www.belkin.com/us/p/P-WSP090/).  These devices have firmware based on an `RTOS` operating systems. Other WeMo devices I've looked at have firmware based on `OWRT` (Open WRT, I think). The ones with the `OWRT` firmware are following the UPnP spec and do work with this PR.

All `RTOS` & `OWRT` devices do send subscription event notifications when their state is changed. It's just that the `RTOS` devices don't send the initial notification when the subscription is first created. That makes it impossible to detect state changes that happen when the subscription drops (without resorting to polling) and also impossible for this PR to confirm that the device is capable of sending state changes (i.e. not blocked by a network device - Firewall/NAT) without causing a state change.

I'm going to look to see if there is a way to get this PR working with the `RTOS` devices later today.

Spectre5:
What a mess!

esev:
@Spectre5 it gets more annoying too: #237 (though in this case I don't think the changes are necessarily out of conformance with the UPnP specifications)

esev:
Another odd thing about the `RTOS` devices is that they don't provide the normal `urn:Belkin:service:basicevent:1` ST value in UPnP/SSDP discovery messages in response to an `M-SEARCH` request containing that same ST value. I believe the response from the `RTOS` devices is out of conformance with the UPnP specifications. #158 was required to allow them to be discovered.

[Specs](https://openconnectivity.org/upnp-specs/UPnP-arch-DeviceArchitecture-v2.0-20200417.pdf) (for M-SEARCH requests with ST headers in the form `urn:domain-name:service:serviceType:ver`):
> Respond once for each matching service type. shall be urn: domain-name:service:serviceType:ver where domain-name (a Vendor Domain Name), serviceType and ver are defined by the UPnP vendor and ver shall contain the version of the service type from the M-SEARCH request. Period characters in the Vendor Domain Name shall be replaced with hyphens in accordance with RFC 2141. 

`M-SEARCH` request sent by pyWeMo (`ST: urn:Belkin:service:basicevent:1`)
```http
M-SEARCH * HTTP/1.1
HOST: 239.255.255.250:1900
MAN: "ssdp:discover"
MX: 2
ST: urn:Belkin:service:basicevent:1
```

Responses:

WSP080 (`ST: upnp:rootdevice`)
```http
HTTP/1.1 200 OK
CACHE-CONTROL: max-age=86400
DATE: Wed, 10 Feb 2021 03:44:42 GMT
EXT:
LOCATION: http://192.168.1.100:49152/setup.xml
OPT: "http://schemas.upnp.org/upnp/1/0"; ns=01
X-User-Agent: redsonic
ST: upnp:rootdevice
USN: uuid:Socket-1_0-SerialNumber::upnp:rootdevice
```

WSP090 (`ST: upnp:rootdevice`)
```http
HTTP/1.1 200 OK
CACHE-CONTROL: max-age=86400
DATE: Wed, 10 Feb 2021 03:41:45 GMT
EXT:
LOCATION: http://192.168.1.100:49152/setup.xml
OPT: "http://schemas.upnp.org/upnp/1/0"; ns=01
X-User-Agent: redsonic
ST: upnp:rootdevice
USN: uuid:OutdoorPlug-1_0-SerialNumber::upnp:rootdevice
```

F7C059, for comparison (`ST: urn:Belkin:service:basicevent:1`)
```http
HTTP/1.1 200 OK
CACHE-CONTROL: max-age=86400
DATE: Sun, 09 Jan 2000 00:16:34 GMT
EXT:
LOCATION: http://192.168.1.100:49157/setup.xml
OPT: "http://schemas.upnp.org/upnp/1/0/"; ns=01
01-NLS: <snip UUID value >
X-User-Agent: redsonic
ST: urn:Belkin:service:basicevent:1
USN: uuid:Dimmer-1_0-SerialNumber::urn:Belkin:service:basicevent:1
```

esev:
I think this PR should be okay to merge now. I had been holding off because this PR hides the issues being fixed in #239. It was easier for me to work on that other PR while this one was not merged.

I don't think there is anything pywemo can do to safely confirm that the RTOS devices are successfully subscribed. The only safe thing to do for these devices is to poll. If Belkin fixes the RTOS devices to conform with the UPnP spec in the future then they should just start working with this PR too. Clients that don't need the additional safety confirmation of seeing the initial event notification can disable polling without checking `is_subscribed`.

PR #238, opened 2021-02-13
Discovery optimizations

esev:
## Description:

- Remove mac/serial/service_types comparisons from `ssdp.scan`. As a side-effect, /setup.xml no longer needs to be fetched in `ssdp.scan`.
- Fix `__eq__` for `UPNPEntry` and remove `entry_in_entries`
- Remove the `DESCRIPTION_CACHE` in `ssdp.scan`. It is no longer required as `ssdp.scan` is no longer fetching /setup.xml
- Avoid a 2nd fetch of /setup.xml in `discover_devices` by calling `device_from_uuid_and_location` directly.
- Deprecate the `mac` argument to the Device constructor .

**Related issue (if applicable): https://github.com/home-assistant/core/issues/45904

## Checklist:

### DIFF
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -182,9 +182,7 @@ class UPNPEntry:
                     tree = et.fromstring(xml or b'')
                     return etree_to_dict(tree).get('root', {})
                 except requests.RequestException:
-                    logging.getLogger(__name__).warning(
-                        "Error fetching description at %s", url
-                    )
+                    LOG.warning("Error fetching description at %s", url)
 
         except et.ParseError:
             # There used to be a log message here to record an error about
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -174,7 +174,10 @@ class UPNPEntry:
     @property
     def description(self):
         """Return the description from the uPnP entry."""
-        url = self.values.get('location', '_NO_LOCATION')
+        url = self.location
+        if not url:
+            return {}
+
         try:
             for _ in range(3):
                 try:
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -240,7 +240,6 @@ class UPNPEntry:
 
 def build_ssdp_request(ssdp_st, ssdp_mx):
     """Build the standard request to send during SSDP discovery."""
-    ssdp_st = ssdp_st or ST
     return "\r\n".join(
         [
             'M-SEARCH * HTTP/1.1',
@@ -254,7 +253,7 @@ def build_ssdp_request(ssdp_st, ssdp_mx):
     ).encode('ascii')
 
 
-def scan(st=None, timeout=DISCOVER_TIMEOUT, max_entries=None, match_udn=None):
+def scan(st=ST, timeout=DISCOVER_TIMEOUT, max_entries=None, match_udn=None):
     """
     Send a message over the network to discover upnp devices.
 
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -179,13 +179,8 @@ class UPNPEntry:
             for _ in range(3):
                 try:
                     xml = requests.get(url, timeout=REQUESTS_TIMEOUT).content
-
-                    if xml is not None:
-                        tree = et.fromstring(xml)
-                        if tree is not None:
-                            return etree_to_dict(tree).get('root', {})
-                    break
-
+                    tree = et.fromstring(xml)
+                    return etree_to_dict(tree).get('root', {})
                 except requests.RequestException:
                     logging.getLogger(__name__).warning(
                         "Error fetching description at %s", url
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -133,8 +133,6 @@ class SSDP:
 class UPNPEntry:
     """Found uPnP entry."""
 
-    DESCRIPTION_CACHE = {'_NO_LOCATION': {}}
-
     def __init__(self, values):
         """Create a UPNPEntry object."""
         self.values = values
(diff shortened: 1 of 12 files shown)


pywemo/ssdp.py
```diff
@@ -345,39 +361,29 @@ def scan(  # noqa: C901
...
-                    serial = None
-                    service_types = []
 
                 # Search for devices
-                if any(i is not None for i in [st, match_mac, match_serial]):
-                    if not entry_in_entries(entry, entries, mac, serial):
-                        if match_mac is not None:
-                            if match_mac == mac:
-                                entries.append(entry)
-                        elif match_serial is not None:
```
esev, line 369: It wasn't clear to me why `elif` is used here. Was that intentional? 

Ex: What should the behavior be if both `match_mac` and `match_serial` are provided? Should `match_serial` be ignored in that case (priority order), should both be required to match (logical AND), or should a match be made if either values match (logical OR)?

Maybe we should warn if two match_ values are being used? I think logical OR makes more sense. WDYT? I can change the new code if needed.

esev: Taking a step back: Could we remove match_mac, match_serial & match_st and just keep match_udn? That would clean-up the code a bit and remove the question above.

The UDN (Unique Device Name) should be unique across all WeMo devices. It looks like it combines the model + serial. It is contained directly in the UPnP response (it does not require fetching /setup.xml) and is also a [required field](https://github.com/pavoni/pywemo/blob/1253cdd32a4c0d63aa628c52cc0420308a1ce322/pywemo/ouimeaux_device/api/xsd/device.xsd#L46) in /setup.xml too.

I don't have the context on why match_mac & match_serial were added. (I suspect match_serial was added for the case where the mac argument to `Device.__init__` is `''`/`None`). But I also don't have the context for why [`mac`](https://github.com/pavoni/pywemo/commit/985b9031144e826f73758ff0e834b291b5107e37#diff-ac30cfeca275251c62af9cc8586687c7ab9af903b503c076c54f813c4e3027e9) is a required argument to `Device.__init__`. It seems odd to me that discovery parses /setup.xml to extract the mac address only to pass it to `Device.__init__` which then parses /setup.xml again. It looks like [`macAddress`](https://github.com/pavoni/pywemo/blob/1253cdd32a4c0d63aa628c52cc0420308a1ce322/pywemo/ouimeaux_device/api/xsd/device.xsd#L47) is not a required field in the schema, so maybe that has something to do with it?

Spectre5: I've had some of the same questions myself, so maybe @pavoni recalls some of the history here?

pavoni: Some of this is before my time (even), but at some point the discovery was integrated with HA’s discovery, which I think passed the mac.

So may be ghosts of how HA worked long ago. At the time I think the only reliable unique identifier that could be used inside HA was the mac, esp when it was thought wemos were changing IP.

As long as it works stand alone and inside HA, fee free to refactor.

Rediscovery has always been shaky, so would be great to fix.

esev: Thanks!

I've made `mac` optional and added a deprecation warning so that we can remove it in the future. I also removed `match_mac` & `match_serial`.

This greatly speeds up `ssdp.scan` since it no longer needs to read setup.xml. All time in `ssdp.scan` is spent ... on scanning. :)

And, since setup.xml no longer needs to be read as part of `ssdp.scan` I've also removed the `DESCRIPTION_CACHE`. IIRC it was only there as an optimization since it's typical that `ssdp.scan` receives multiple responses from the same device.

pywemo/ssdp.py
```diff
@@ -164,47 +162,42 @@ def location(self):
...
-        if url not in UPNPEntry.DESCRIPTION_CACHE:
-            try:
-                for _ in range(3):
-                    try:
-                        xml = requests.get(
-                            url, timeout=REQUESTS_TIMEOUT
-                        ).content
+                    if xml is not None:
+                        tree = et.fromstring(xml)
+                        if tree is not None:
```
Spectre5: Can `fromstring` even return None?  I think it returns a valid object or an exception in all cases

Spectre5:
I'm really liking these changes, it cleans up a lot of "ghost" code and is overall more clean, let alone the improved scan response speeds that @esev is already seeing with it.  I'm going to finish reviewing it either later tonight or tomorrow evening.

pywemo/ssdp.py
esev: Good point. We don't need to check the output of `requests.get` either. If it gets an empty response then lxml will also fail and the code will handle the exception below.

```
>>> et.fromstring('')
  File "src/lxml/etree.pyx", line 3237, in lxml.etree.fromstring
  File "src/lxml/parser.pxi", line 1896, in lxml.etree._parseMemoryDocument
  File "src/lxml/parser.pxi", line 1777, in lxml.etree._parseDoc
  File "src/lxml/parser.pxi", line 1082, in lxml.etree._BaseParser._parseUnicodeDoc
  File "src/lxml/parser.pxi", line 615, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 725, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 654, in lxml.etree._raiseParseError
  File "<string>", line 1
lxml.etree.XMLSyntaxError: Document is empty, line 1, column 1
```

Spectre5: Right, as long as `get().content` cannot actually return `None`, since `None` raises a `ValueError` if passed to `fromstring`.  That said, I'm fairly certain `content` cannot be None

esev: Ah bummer. Looks like there is an edge case where `content` can be `None`.

https://github.com/psf/requests/blob/8c211a96cdbe9fe320d63d9e1ae15c5c07e179f8/requests/models.py#L828-L831

Fixed.

Spectre5: Always good to check the source - I guess I've just never *seen* it be None!

Spectre5: But great fix for that edge case, no need for the `if` statement

esev:
Rebased to the master branch after e05aefb

Spectre5 (approved):
Everything looks good to me and works for me in some in-home tests.

PR #239, opened 2021-02-15
Add retries for individual SOAP actions

esev:
## Description: 

Add retries for individual SOAP actions.

WeMo devices with weak wifi connections often get stuck in the unavailable state in Home Assistant for a few minutes at a time. The main cause for this is because the Device enters probing/rediscovery as soon as a SOAP action request fails.

**Before**
The graph shows the number of times the pywemo library timed-out waiting for the WeMo device to respond over a 10 minute period. Each color represents a different device.
Screenshot 2021-02-08 1 38 33 PM

And this graph shows the amount of time spent attempting to reconnect to the device when the SOAP action has a timeout. (Gaps in the graph are periods of time where no reconnects happened)
Screenshot 2021-02-11 1 47 59 PM

**After**
This PR adds 6 retries to the SOAP actions. These retries happen before attempting to reconnect, allowing the action to complete in the majority of cases without triggering a reconnection.

With this PR, the SOAP action completes most of the time. As with the previous graph the colors in this one represent individual devices. Only [one device](https://github.com/home-assistant/core/issues/45904#issuecomment-778912833) on my wifi still has issues now.
Screenshot 2021-02-15 at 8 32 41 AM

And in 99 percentile of the cases, including the one flaky device, the request completes in less than 5 seconds, occasionally requiring 1 retry. (The median over this same period is about 0.11 seconds.)
Screenshot 2021-02-15 at 8 34 57 AM

There are a few SOAP actions that are intended to take longer than the default timeout. I've added a separate timeout for those to allow them to still work correctly (see: `Action.soap_action_timeout_override`). I've also added a `pywemo_timeout` kwarg to allow clients to override the timeouts on individual requests. [I would have found this helpful when initially setting up my Bridge device]

In addition to retrying the SOAP actions, this PR also causes the Device description request for /setup.xml to be retried as well as retrying the fetch of all the service description xml files too.

Note that the urllib3 dependency is not new. The requests library was already depending on it. I simply copied the version requirements from the requests library and added urllib3 to pyproject.toml. 

I described in the comments for the `Session` class why urllib3 is being used. The requests library sets [`preload_content=False`](https://github.com/psf/requests/search?q=preload_content), making the fetch of the body/content happen outside of the urllib3 retries logic. Because of this, there doesn't appear to be any way to have the requests library automatically retry when the fetch of the body times-out. An alternative could be to re-implement the urllib3 retry logic in pywemo, but I'm choosing to re-use rather than re-implement.

Finally, after this PR the URL/host/port are all contained in a single location inside the `Session` instance. All other uses of the URL/host/port reference it from that single location. This makes it possible for the reconnection logic to update the device location by just setting `device.session.url`. After this PR, and #238, are merged I'll send another PR to just update `self.session.url`, rather than creating a new device and replacing the inner `self.__dict__` of properties. There is a test case for this in test_subscribe.py.

**Related issue (if applicable):** https://github.com/home-assistant/core/issues/45904

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/api/service.py
+++ b/pywemo/ouimeaux_device/api/service.py
@@ -1,14 +1,15 @@
 """Representation of Services and Actions for WeMo devices."""
 # flake8: noqa E501
 import logging
+from urllib.parse import urljoin, urlparse
 
 import requests
+import urllib3
 from lxml import etree as et
 
 from .xsd import service as serviceParser
 
 LOG = logging.getLogger(__name__)
-MAX_RETRIES = 3
 REQUESTS_TIMEOUT = 10
 
 REQUEST_TEMPLATE = """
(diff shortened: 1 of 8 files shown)


Spectre5:
If listing urllib3 in the pyproject.toml, then it should probably be added to the README as well, for consistency with the other dependencies:
`pyWeMo depends on Python packages: requests, ifaddr, lxml`

pywemo/ouimeaux_device/api/service.py
```diff
@@ -29,73 +30,210 @@ class ActionException(Exception):
...
 class Action:
     """Representation of an Action for a WeMo device."""
 
-    def __init__(self, device, service, action_config):
+    # A few actions take longer than the default timeout. Override the default
+    # timeout value for those actions.
+    soap_action_timeout_override = {
+        "urn:Belkin:service:bridge:1#AddDevice": 30,
+        "urn:Belkin:service:bridge:1#OpenNetwork": 30,
+        "urn:Belkin:service:WiFiSetup:1#GetApList": 10,
```
Spectre5, line 167: 10 is the default, is this one supposed to be 30 too?

Spectre5: Nevermind, this is really for the `PoolManager` in the `Session` class, which has a default of 3

esev: I changed the default for SOAP action requests to 3 seconds (`Session.timeout`). I've kept REQUESTS_TIMEOUT at 10 seconds as not all of the code that uses the requests library has retries enabled. In my tests `urn:Belkin:service:WiFiSetup:1#GetApList` completes in just over 4 seconds.

Spectre5 (approved):
I've looked through it all and don't have any issues.  A little unfortunate about switching between requests and urllib3, but I see why it was done too.  It would look a little more like a hack to force requests to use a timeout for the pool manager that is currently done here.  Would it make sense to add an extra attribute to the urllib3 response of `response.content = response.data` just to make that part look a little more like requests, so the handful of `response.contents` would stay as that instead of changing to `response.data`?

esev:
> Would it make sense to add an extra attribute to the urllib3 response of response.content = response.data just to make that part look a little more like requests, so the handful of response.contents would stay as that instead of changing to response.data?

Makes sense to me.

PR #240, opened 2021-02-17
Combine all pyWeMo exceptions in a single file

esev:
## Description:

Make it easier for clients handle exceptions by allowing them to be imported from one location.

```python
try:
    device.on()
except pywemo.PyWeMoException:
    pass
```

or

```python
try:
    device.on()
except pywemo.exceptions.ActionException:
    pass
```

## Checklist:

### DIFF
--- a/pywemo/__init__.py
+++ b/pywemo/__init__.py
@@ -1,15 +1,16 @@
 """Lightweight Python module to discover and control WeMo devices."""
+# flake8: noqa F401
 
-from .discovery import discover_devices  # noqa F401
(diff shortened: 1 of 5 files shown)


esev:
We don't need to merge this. Just looking for your thoughts @pavoni & @Spectre5. Do you think this is helpful?

Spectre5:
I'd be in support of this.  I think it makes sense to have a base underlying exception for the entire library.  And putting all exceptions in a separate file is probably best practice too.

PR #241, opened 2021-02-17
Reconnect without creating a new device

esev:
## Description:

Reconnect without creating a new device.

I've also removed the retries logic from `_reconnect_with_device_by_discovery`. Since the SOAP requests are retried, and `ssdp.scan` has its own retry logic, I didn't think we needed another set of retries in `_reconnect_with_device_by_discovery` too.

**Related issue (if applicable):** https://github.com/home-assistant/core/issues/45904

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -3,6 +3,7 @@
 import base64
 import logging
 import subprocess
+import threading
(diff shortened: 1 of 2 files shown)


pavoni:
LGTM.

Spectre5 (approved):
Also looks good

PR #242, opened 2021-02-18
Clients can use HTTPException

esev:
## Description:

pyWeMo clients shouldn't need to know the library used to make http requests to be able to handle exceptions. This PR adds a `pywemo.exceptions.HTTPException` that is raised any time a client visible http request fails. This allows clients to interact with the library and expect all exceptions will map to the top-level `pywemo.PyWeMoException`.

```python
try:
    device = pywemo.discovery.device_from_description(url)
except pywemo.PyWeMoException:
    pass
```

I only added this exception in places where the exception is visible to the client. The exceptions in `subscribe.py`, for example, are only visible internally in pyWeMo.

## Checklist:

### DIFF
--- a/pywemo/discovery.py
+++ b/pywemo/discovery.py
@@ -7,8 +7,9 @@ from socket import gaierror, gethostbyname
 import requests
 
 from . import ssdp
+from .exceptions import ActionException, HTTPException
 from .ouimeaux_device import UnsupportedDevice, probe_wemo
-from .ouimeaux_device.api.service import REQUESTS_TIMEOUT, ActionException
+from .ouimeaux_device.api.service import REQUESTS_TIMEOUT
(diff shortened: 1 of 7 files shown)


pavoni:
Also looks good.

Spectre5 (approved):
Looks good to me

Spectre5:
I know there is a way to make coveralls have some allowable amount of decrease for cases like this.  We should probably allow something like a 2% or 3% decrease in coverage for smaller PR's and simple changes.

Spectre5:
Ah, looks like that would be something @pavoni needs to change on the coveralls website under settings.

pavoni:
> Ah, looks like that would be something @pavoni needs to change on the coveralls website under settings.

Done (I think!).

esev:
> (I think!).

Indeed :)  Thanks!

PR #250, opened 2021-02-20
add pre-commit and some hooks

### LINKED ISSUE #199: [Development] Add a pre-commit hook
Making this issue/discussion mainly as a reminder.  We may want to look into running the various checks from the `build.sh` file as a pre-commit hook.  It could be installed as part of the build.sh file as well (or a new, separate file that installs poetry and the pre-commit hook only).

Spectre5:
## Description:
This adds several pre-commit hooks using the pre-commit python package.  We may want a ~short~ (became long!) discussion on this first.  First, 2 quick notes:

- Poetry normally recommends that you install it outside of your local project, see their instructions [here](https://github.com/python-poetry/poetry#installation).  I don't really love the idea of installing it with some special script, but I do understand the purpose of installing it outside of your local virtual environment, mainly that it can then be used to create the virtual environment itself.  For other projects, I've installed it with `pipx` so that it is globally accessible and in an isolated environment.  PyWeMo gets around that by using a few bash functions, `assertPython` and `enterVenv`, then installing Poetry into the created environment and using it.

- Pre-commit generally recommends that you point it at a git repository, that hosts a `.pre-commit-hooks.yaml` file at the top level of the directory.  For example, see [here](https://pre-commit.com/#2-add-a-pre-commit-configuration).  Pre-commit then manages an independent virtual environment for each "hook" that you install.  Personally, I don't like this method for a python project as it seems pointless to create extra virtual environment for tools that you already install into your project development virtual environment!  Further, now you need to manage the version of the tools you use in two places, the `pyproject.toml` file and your `.pre-commit-config.yaml` file, because occasionally different versions of these tools are not fully compatible.  For example, `black` version `19.10b0` and the next version of `20.8b1` would reformat some parts of code slightly differently.  This is one reason that I've pinned `black` and `isort` in our `pyproject.toml` file, so that we ensure we all format with the same version.

My solution to the second bullet is to use the [`system`](https://pre-commit.com/#system) "escape" language and then run the tools directly with `poetry`, which will automatically use the projects virtual environment to run whatever tool you request.  The catch here, is that you must have `poetry` installed somewhere outside of the virtual environment for this to work (see bullet 1), as pre-commit does not activate your project environment, so if you don't have `poetry` installed somewhere globally accessible on your `$PATH`, then pre-commit isn't going to find it (unless you commit with your virtual environment already activated, I suppose).

So, we just need to all agree on which path to take.  I can see good arguments for each way.

- Use pre-commit as it "recommends" and then just manage the versions of each tool in two locations (and probably pin each versions to be sure they are the same)
- Use what I've commited here, but make it clear that there is a requirement that `poetry` be installed outside of the virtual environment.  If we go this route, we could potentially just use `poetry` to also create the virtual environment for the project.

As for the settings themselves, I pretty much copied and edited the settings from each of the tools github repositories `.pre-commit-hooks.yaml` file, so that is where I got the various `types`, `types_or`, and `require_serial` options from.

I do not include pytest in this pre-commit hooks, as I assume we don't want to wait for all tests to run on every commit.  I tested committing a few "bad" files, like a bad `README.rst` and the hook did catch and reject it.

Sorry this is getting a bit longer than I'd planned, but 2 more points to bring up:

- We could potentially just run the `build.sh` script as a pre-commit hook instead (which is what @esev did?),  but then we would run the installations and tests as well.  Or we could break up the `build.sh` file into two, where only one would be installed as a pre-commit hook.  `pre-commit` seems to be the standard though, so I went that path.
- We could replace a bit of the `build.sh` file with just `pre-commit run --all-files`, but I personally prefer them all called separately in the `build.sh` files myself.  Plus then it would only check with `black` and `isort`, not actually format.

**Related issue (if applicable):** fixes #199

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,49 @@
+fail_fast: true
+repos:
+  - repo: local
+    hooks:
+      - id: poetry-check
+        name: poetry-check
+        language: system
+        entry: poetry check
+        pass_filenames: false
+        types: [file]
+        files: ^(pyproject.toml|poetry.lock)$
+
+      - id: isort
+        name: isort
+        language: system
+        entry: poetry run isort
+        args: ['--check-only', '--filter-files']
+        require_serial: true
+        types_or: [cython, pyi, python]
+
+      - id: black
+        name: black
+        language: system
+        entry: poetry run black
+        args: ['--check']
+        require_serial: true
+        types_or: [python, pyi]
+
+      - id: flake8
+        name: flake8
+        language: system
+        entry: poetry run flake8
+        require_serial: true
+        types: [python]
+
+      - id: pylint
+        name: pylint
+        language: system
+        entry: poetry run pylint
+        args: ['--disable=fixme']
+        require_serial: true
+        types: [python]
+        exclude: tests/
+
+      - id: rstcheck
+        name: rstcheck
+        language: system
+        entry: poetry run rstcheck
+        types: [rst]
--- a/poetry.lock
+++ b/poetry.lock
@@ -72,6 +72,14 @@ optional = false
 python-versions = "*"
 
 [[package]]
+name = "cfgv"
+version = "3.2.0"
+description = "Validate configuration and produce human readable error messages."
+category = "dev"
+optional = false
+python-versions = ">=3.6.1"
+
+[[package]]
 name = "chardet"
 version = "4.0.0"
 description = "Universal encoding detector for Python 2 and 3"
@@ -123,6 +131,14 @@ requests = ">=1.0.0"
 yaml = ["PyYAML (>=3.10)"]
 
 [[package]]
+name = "distlib"
+version = "0.3.1"
+description = "Distribution utilities"
+category = "dev"
+optional = false
+python-versions = "*"
+
+[[package]]
 name = "docopt"
 version = "0.6.2"
 description = "Pythonic argument parser, that will make you smile"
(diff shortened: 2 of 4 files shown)


esev (approved):
I think this works great as-is with my workflow.

I don't have poetry installed on my host, but the hook still works well if I enter my venv first. That's my normal workflow anyway, since I'm typically running the tests manually before I do a commit. I get the error message ```Executable `poetry` not found``` if I run it without entering my venv. That's a good reminder if I had forgotten to enter the venv first.

It looks like what you have here works pretty well on its own. I don't think there needs to be more integration between build.sh and the hooks, other than what you have already have setup. Maybe in the future we could add a script that just sets-up the venv and the hooks for getting started initially, but running build.sh works well for this right now.

Spectre5:
One more comment on this - another option would be to call the scripts from within the virtual environment directly without using `poetry`.  This wouldn't generally be portable, but since `build.sh` always installs the project venv into `.venv`, it would work.

So, for example, we could change `poetry run black` to `.venv/bin/black`

However, if anyone tries this with Windows, this will cause an issue - on Windows the virtual environments install things into `.venv/Scripts` instead of `.venv/bin` (which was a terrible design decision).  Although I don't use Windows, I think using `poetry` as I current do is probably better so it is slightly more portable.  If people have `poetry` globally accessible then it'll just work and if not, they'll get an error as @esev has found.

Actually, a Windows user would already have an issue using `build.sh` anyway as it uses bin instead of Scripts directly in the `enterVenv` function:
 
```bash
    ...
    source "$VENV_DIR/bin/activate"
    ...
```

Edit: just to clarify - when mentioning Windows above I'm talking about using `git-bash` or similar, obviously `build.sh` wouldn't work with `cmd` or powershell.

PR #252, opened 2021-02-21
make port optional to setup_url_for_address

Spectre5:
## Description:
Make the `port` argument to `setup_url_for_address` optional by assigning a default value of `None`.  I'd think the vast majority of use cases would provide `None` for the argument anyway so that `pywemo` can probe for the port.

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ On some networks discovery doesn't work reliably, in that case if you can find t
 .. code-block:: python
 
     >>> import pywemo
-    >>> url = pywemo.setup_url_for_address("192.168.1.192", None)
+    >>> url = pywemo.setup_url_for_address("192.168.1.192")
     >>> print(url)
     http://192.168.1.192:49153/setup.xml
     >>> device = pywemo.discovery.device_from_description(url)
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -25,7 +25,7 @@ On some networks discovery doesn't work reliably, in that case if you can find t
 .. code-block:: python
 
     >>> import pywemo
-    >>> url = pywemo.setup_url_for_address("192.168.1.192", None)
+    >>> url = pywemo.setup_url_for_address("192.168.1.192")
```
esev, line 31: WDYT about changing the documentation after the next release? That way the examples match what folks are able to download from pypi.

Spectre5:
Ya I considered that issue too.  Maybe we just sit on this PR until that time and leave it open as a reminder?

README.rst
Spectre5: Oops, see reply in main comments.

esev:
> Maybe we just sit on this PR until that time and leave it open as a reminder?

That sounds good. Let's do the change to discovery.py in another PR so it can be merged now. Then keep this open for after the next release.

PR #225, opened 2021-01-30
Update license wording.

pavoni:
## Description:
Change the license wording to cover changes to the `ouimeaux` directory 

**Related issue (if applicable):** closes #222

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] There is no commented out code in this PR.

PR #255, opened 2021-03-07
clarifications to the README, mainly setup/reset

Spectre5:
## Description:
I've re-written parts of the README, mainly the text in the setup and reset methods to hopefully make it more clear and also link to @esev's guide for blinking light issues if the device has no connection to the internet.  I've put all of the code stuff to have two back-ticks so that it gets formatted correctly.  I've seen other reStructuredText files do this for README files, so I believe that PyPI will not have an issue with that (I once tried using a :code: block to do it but PyPI complained --- I think this way works though).

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ For advanced usage, the ``device.explain()`` method will print all known actions
 If discovery doesn't work on your network
 -----------------------------------------
 Automatic discovery may not work reliably on some networks.
-In that case, you can find the device based on an IP or hostname:
+In that case, you can use the device with an IP or hostname:
 
 .. code-block:: python
 
@@ -65,7 +65,7 @@ Setup
 
 Device setup is through the ``setup`` method, which has two required arguments: ``ssid`` and ``password``.
 The user must first connect to the devices locally broadcast access point, which typically starts with "WeMo.", and then discover the device there.
-Once done, pass the desired SSID and password (AES encryption only) to the ``setup`` method to connect it to your wifi network.
+Once done, pass the desired SSID and password (WPA2/AES encryption only) to the ``setup`` method to connect it to your wifi network.
 
 ``device.setup(ssid='wifi_name', password='special_secret')``
 
@@ -75,7 +75,7 @@ A few important notes:
 - WeMo requires a password of at least 8 characters long.
 - For a WeMo without internet access, see `this guide <https://github.com/pavoni/pywemo/wiki/WeMo-Cloud#disconnecting-from-the-cloud>`_ to stop any blinking lights.
 - If connecting to an open network, the password argument is ignored and you can provide anything, e.g. ``password=None``.
-- If connecting to an AES-encrypted network, OpenSSL is used to encrypt the password by the ``pywemo`` library.
+- If connecting to a WPA2/AES-encrypted network, OpenSSL is used to encrypt the password by the ``pywemo`` library.
   It must be installed and available on your ``PATH`` via calling ``openssl`` from a terminal or command prompt.
 
 Firmware Warning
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -18,9 +18,12 @@ How to use
 
     >>> devices[0].toggle()
 
+For advanced usage, the ``device.explain()`` method will print all known actions that the device reports to PyWeMo.
+
 If discovery doesn't work on your network
 -----------------------------------------
-On some networks discovery doesn't work reliably, in that case if you can find the ip address of your Wemo device you can use the following code.
+Automatic discovery may not work reliably on some networks.
+In that case, you can find the device based on an IP or hostname:
```
esev: Instead of `you can find the device based on` how about `you can use the device with`?

README.rst
```diff
@@ -32,58 +35,83 @@ On some networks discovery doesn't work reliably, in that case if you can find t
...
-OpenSSL is used to encrypt the password by the pywemo library.
-It must be installed and available on the path via calling `openssl` with a terminal (or command prompt, if on Windows).
-This is not required if connecting the device to an open network, since that requires no password, although an open network certainly isn't recommended.
+A few important notes:
+
+- Not all devices are supported for setup.
+- WeMo requires a password of at least 8 characters long.
+- For a WeMo without internet access, see `this guide <https://github.com/pavoni/pywemo/wiki/WeMo-Cloud#disconnecting-from-the-cloud>`_ to stop any blinking lights.
+- If connecting to an open network, the password argument is ignored and you can provide anything, e.g. ``password=None``.
+- If connecting to an AES-encrypted network, OpenSSL is used to encrypt the password by the ``pywemo`` library.
```
esev: I think AES (and the password length) are requirements of WPA2. I feel folks are more likely to understand the term WPA2 than AES-encrypted. WDYT?

README.rst
```diff
@@ -32,58 +35,83 @@ On some networks discovery doesn't work reliably, in that case if you can find t
...
+    . .venv/bin/activate
+
 History
 -------
 This started as a stripped down version of `ouimeaux <https://github.com/iancmcc/ouimeaux>`_, but has since taken its own path.
 
 License
 -------
-Some of the code in `pywemo/ouimeaux_device` was originally written by by Ian McCracken (and is copyright). It is released under the BSD license.
+Some of the code in `pywemo/ouimeaux_device` was originally written, copyright, and released under the BSD license by Ian McCracken.
```
esev: Mind if we switch the sentence ordering a bit?
- The code in `pywemo/ouimeaux_device` is released under the BSD LICENSE file in that directory. Some of this code was originally written and copyright by Ian McCracken.

I'm thinking that might keep closer with the wording prior to #225 while also clarifying the copyright.

README.rst
Spectre5: Updated.

README.rst
Spectre5: Sure, I'll change AES to WPA2, you're right that most people will be more familiar with that terminology.   As for the password length, I have no idea.  You're probably right, I've never tried a shorter password for my network.  Should I just remove that bullet then since that will always be satisfied if they are using WPA2?

README.rst
Spectre5: Sure, makes sense.  And more updates to this may come when you do the contributing guide and CLA to make it all more clear.

README.rst
Spectre5: Ya, length 8 is a minimum for WPA2.  I'll just remove that bullet

Spectre5:
I'm not sure if we should specify which devices we know that the `Setup` doesn't currently work with?  Maybe that kind of information is best put in the wiki though

README.rst
esev: >Should I just remove that bullet then since that will always be satisfied if they are using WPA2?

Removing that part SGTM. I've been wondering if the `ShortPassword` exception is really an indicator of `IncorrectPassword`. I'd need to do more testing to verify that.

esev:
> I'm not sure if we should specify which devices we know that the Setup doesn't currently work with? Maybe that kind of information is best put in the wiki though

I think having a _device quirks_ (or similar) wiki page to describe all this would be helpful, yes.

esev (approved):

PR #280, opened 2021-07-24
Add info on getting IP address manually when connected directly to WeMo

### LINKED ISSUE #271: Suggestion for Setup section in README
Big fan of pywemo, thanks for supporting it! I recently changed my network name/password and had to reconfigure my devices, I put them in discovery mode and connected to it directly (using the "Wemo" network).

It might be helpful to mention you can run `arp -a` to get the ip address used to connect.
```
$ arp -a
_gateway (10.22.22.1) at [REMOVED] [ether]
```
and then in python:
```
>>> import pywemo
>>> url = pywemo.setup_url_for_address("10.22.22.1")
>>> device = pywemo.discovery.device_from_description(url)
>>> print(device)
<WeMo Switch "Wemo Mini">
>>> device.setup(ssid='MY SSID', password='MY NETWORK PASSWORD')
('1', 'success')
```

Thanks again! Let me know and I can PR the README if you want!

GarlicToum:
## Description:
I ran into a situation where the `discover_devices` didn't work on my computer when connected directly to the WeMo device. I found if you do an `arp -a` you can get the ip address and then connect manually. Also, if ya'll don't wont the extra python code block, feel free to remove it.

fixes #271 

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,7 @@ For advanced usage, the ``device.explain()`` method will print all known actions
 If discovery doesn't work on your network
 -----------------------------------------
 Automatic discovery may not work reliably on some networks.
-In that case, you can use the device with an IP or hostname from your network.
+In that case, you can use the device with an IP or hostname:
 
 .. code-block:: python
 
--- a/README.rst
+++ b/README.rst
@@ -44,7 +44,7 @@ The ``setup_url_for_address`` function will lookup a hostname and provide a suit
 If the WeMo device is not on your network, you can also connect to it directly.
 After connecting, if the ``pywemo.discover_devices()`` doesn't work, you can get the IP Address by running an ``arp -a`` and use that in ``pywemo.setup_url_for_address``:
 
-.. code-block:: bash
+.. code-block::
 
     $ arp -a
     _gateway (10.22.22.1) at [MAC ADDRESS REMOVED] [ether]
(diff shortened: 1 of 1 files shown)


esev:
Thank you!

esev:
Hmm. Looks like the test failed, though the page looks fine when I view it. 

@Spectre5 could you take look when you have a chance? If I merge it as-is I think the tests will fail in the main branch.

README.rst
```diff
@@ -41,6 +41,26 @@ In addition, recent versions of the WeMo firmware may not accept connections fro
 
 The ``setup_url_for_address`` function will lookup a hostname and provide a suitable ``url`` with an IP address.
 
+If the WeMo device is not on your network, you can also connect to it directly.
+After connecting, if the ``pywemo.discover_devices()`` doesn't work, you can get the IP Address by running an ``arp -a`` and use that in ``pywemo.setup_url_for_address``:
+
+.. code-block:: bash
```
esev: I think removing 'bash' here might fix the test. I don't know why the test fails for the bash output but succeeds for Python output.

GarlicToum: Oh weird, I've gone and removed it.

esev:
That did the trick. Thanks again!

Spectre5:
You've been busy!  Sorry for not responding sooner but glad you found the change to make.

Spectre5:
For what it's worth, instead of `.. code-block:: bash` we should have used `.. code-block:: shell-session` or `.. code-block:: console` in the original PR.  `bash` is for just bash commands (inputs only), `shell-session` or `console` is for using `$ ` for inputs and then having outputs after (like the `>>>` for python).

PR #282, opened 2021-07-28
Don't log as exception if it's a caught exception.

balloob:
## Description:
This OS error is caught and reported accordingly. There is no need to print a stack trace.

## Checklist:
  - [ ] There is no commented out code in this PR.
  - [ ] I have read and agree to the [CONTRIBUTING document](


esev:
Related to https://github.com/home-assistant/core/issues/52853?

Thank you!

PR #290, opened 2021-08-04
Raise exception for SOAP Fault errors

esev:
## Description:

With this PR SOAP responses containing a device error will cause an exception to be raised. Prior to this PR, these errors were silently ignored when calling `get_state` (the internal `_state` variable was set to `0`) and the caller had no way of knowing if the call failed. This PR adds a new SOAPFault exception that is a child class of ActionException and is raised when a SOAP response message indicates a device failure.

I discovered this via Home Assistant. One of my devices wasn't controllable but there was no indication in the UI that it had failed. It was failing and was returning a SOAP Fault response. Raising the SOAPFault exception allows the client, HA in this case, to be aware of the failure and update the UI accordingly. In this case, making the failing device unavailable in the UI.

**Related issue (if applicable):** 

## Checklist:

### DIFF
--- a/pywemo/exceptions.py
+++ b/pywemo/exceptions.py
@@ -17,7 +17,7 @@ class SOAPFault(ActionException):
     error_code: str = ""
     error_description: str = ""
 
-    def __init__(self, message, fault_element=None):
+    def __init__(self, message="", fault_element=None):
         """Initialize from a SOAP Fault lxml.etree Element."""
         details = ""
         if fault_element is not None:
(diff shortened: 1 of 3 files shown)


Spectre5:
Just curious, why did your device have a SOAP error?  How do you fix it?  Is it just caused by poor wifi connection or something else?

esev:
I actually have no idea how it got into that state. I was kind of surprised that it wasn't returning http (50x) errors, as that's normally what I've seen. Fixed it by restarting the device.

It was puzzling because I have an automation in Home Assistant to turn this light on automatically based on a motion detector. The light wasn't coming on when there was motion. And in Home Assistant I couldn't control it. Yet it was still showing as available in Home Assistant. I thought something broken on the HA side. But what was really happening was the state changes from HA were not happening and the SOAP error was being ignored.

PR #295, opened 2021-11-13
Add Python 3.10 build check

esev:
## Description:
Add Python 3.10 build check.

In the [release notes](https://www.home-assistant.io/blog/2021/11/03/release-202111/#breaking-changes) for Home Assistant 2021.11.0 there was mention of adding support for Python 3.10 and deprecating 3.8. This PR is aimed toward making sure pyWeMo supports Python 3.10.

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.7, 3.9]
+        python-version: ['3.7', '3.9', '3.10']
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/poetry.lock
+++ b/poetry.lock
@@ -410,17 +410,18 @@ test = ["appdirs (==1.4.4)", "pytest (>=6)", "pytest-cov (>=2.7)", "pytest-mock 
 
 [[package]]
 name = "pluggy"
-version = "0.13.1"
+version = "1.0.0"
 description = "plugin and hook calling mechanisms for python"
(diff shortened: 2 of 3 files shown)


Spectre5:
I think the way GitHub actions works, you may need to put 3.10 in quotes.  Seems it is getting parsed as a float.

esev:
> I think the way GitHub actions works, you may need to put 3.10 in quotes. Seems it is getting parsed as a float.

Ah ha! Thanks for taking a look!

Spectre5:
I think this time you need to re-lock everything?  I'd probably just delete the lock file and have poetry re-create it.

Spectre5:
Poetry uses the lock file when present, so this didn't actually update the version of pytest that poetry uses.  See this line in latest build:

```
 Installing pytest (6.2.2)
```

esev:
> Poetry uses the lock file when present, so this didn't actually update the version of pytest that poetry uses. See this line in latest build:
> 
> ```
>  Installing pytest (6.2.2)
> ```

+1 - That update didn't build because I had a merge conflict. Should be resolved now.

Spectre5:
Ya, just saw it after posting.  I thought I only saw the update in pyproject.toml without the lock, but I guess I'd just overlooked it!  Anyway, looks good now  :+1:

esev:
> Ya, just saw it after posting. I thought I only saw the update in pyproject.toml without the lock, but I guess I'd just overlooked it! Anyway, looks good now 👍

Thanks again for taking a look so quickly!

PR #348, opened 2022-06-05
Add configuration to enable callbacks in containers.

pumbaasdad:
## Description:

I run home assistant in a container without using host networking.  As a result, when pyWeMo subscribes for callbacks, it provides the IP address of the container.  Obviously the wemo device can't access this address, so push notifications are not received.

This PR reads two environment variables that can be used to allow callbacks to work in containers that do not use host networking:

 * `PYWEMO_HTTP_SERVER_PORT` is used to specify the port on which the HTTP server should be started.  If this variable has a value, only that port will be tried.  Although it's extremely likely that port 8989 will be available within a container, this provides a mechanism to ensure that you know which container port to bind to when starting the container.
 * `PYWEMO_CALLBACK_ADDRESS` is used to specify the IP and port that should be provided to the wemo device for callbacks.  This can be set to the host machine's IP address and the port on which its listening.

I did consider adding these values as optional constructor parameters to the `SubscriptionRegistry`.  If I had went down that path, it would have meant that any application that wants to allow pyWeMo to be used in a container would have to provide a way of configuring these values.  By using environment variables, application authors don't have to worry about any of this configuration, and the onus is put on the person starting the container.  Please let me know if you think the constructor approach is more appropriate and I'm happy to go down that path.

## Checklist:

### DIFF
--- a/pywemo/subscribe.py
+++ b/pywemo/subscribe.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import collections
 import logging
+import os
 import sched
 import threading
 import time
@@ -19,7 +20,7 @@ from .ouimeaux_device.api.long_press import VIRTUAL_DEVICE_UDN
 from .ouimeaux_device.api.service import REQUESTS_TIMEOUT
 from .ouimeaux_device.dimmer import DimmerV2
 from .ouimeaux_device.insight import Insight
-from .util import get_ip_address
+from .util import get_callback_address
 
 # Subscription event types.
 EVENT_TYPE_BINARY_STATE = "BinaryState"
@@ -157,8 +158,12 @@ class Subscription:
         if self.subscription_id:  # Renew existing subscription.
             headers = {'SID': self.subscription_id}
         else:  # Start a new subscription.
-            host = get_ip_address(host=self.device.host)
-            callback = f'<http://{host}:{self.callback_port}{self.path}>'
+            callback_address = get_callback_address(
+                host=self.device.host,
+                port=self.callback_port,
+            )
+
+            callback = f'<http://{callback_address}{self.path}>'
             headers = {'CALLBACK': callback, 'NT': 'upnp:event'}
         headers['TIMEOUT'] = f'Second-{self.default_timeout_seconds}'
         return requests.request(
@@ -239,15 +244,23 @@ class HTTPServer(ThreadingHTTPServer):
     outer: SubscriptionRegistry
 
 
-def _start_server() -> HTTPServer | None:
+def _start_server(port: int | None) -> HTTPServer:
     """Find a valid open port and start the HTTP server."""
-    for i in range(0, 128):
-        port = 8989 + i
+    requested_port = port or os.getenv('PYWEMO_HTTP_SERVER_PORT')
+    if requested_port is not None:
+        start_port = int(requested_port)
+        ports_to_check = 1
+    else:
+        start_port = 8989
+        ports_to_check = 128
+
+    for i in range(0, ports_to_check):
+        port = start_port + i
         try:
             return HTTPServer(('', port), RequestHandler)
-        except OSError:
-            continue
-    return None
+        except OSError as error:
+            last_error = error
+    raise last_error
 
 
 def _cancel_events(
@@ -420,7 +433,7 @@ class SubscriptionRegistry:
         'insight',
     )
 
-    def __init__(self) -> None:
+    def __init__(self, requested_port: int | None = None) -> None:
         """Create the subscription registry object."""
         self.devices: dict[str, Device] = {}
         self._callbacks: dict[
@@ -440,6 +453,7 @@ class SubscriptionRegistry:
 
         self._http_thread: threading.Thread | None = None
         self._httpd: HTTPServer | None = None
+        self._requested_port: int | None = requested_port
 
     @property
     def port(self) -> int:
@@ -581,7 +595,7 @@ class SubscriptionRegistry:
 
     def start(self) -> None:
         """Start the subscription registry."""
-        self._httpd = _start_server()
+        self._httpd = _start_server(self._requested_port)
         if self._httpd is None:
             raise SubscriptionRegistryFailed(
                 'Unable to bind a port for listening'
--- a/pywemo/util.py
+++ b/pywemo/util.py
@@ -1,6 +1,7 @@
 """Miscellaneous utility functions."""
 from __future__ import annotations
 
+import os
 import socket
 import warnings
 from collections import defaultdict
@@ -69,12 +70,16 @@ def interface_addresses() -> list[str]:
     return addresses
 
 
-def get_ip_address(host: str = '1.2.3.4') -> str | None:
-    """Return IP from hostname or IP."""
+def get_callback_address(host: str, port: int) -> str | None:
+    """Return IP address & port used by devices to send event notifications."""
+    pywemo_callback_address = os.getenv('PYWEMO_CALLBACK_ADDRESS')
+    if pywemo_callback_address is not None:
+        return pywemo_callback_address
+
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     try:
         sock.connect((host, 9))
-        return cast(str, sock.getsockname()[0])
+        return f'{cast(str, sock.getsockname()[0])}:{port}'
     except OSError:
         return None
     finally:
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -15,7 +15,7 @@ from lxml import etree as et
 
 from .ouimeaux_device.api.long_press import VIRTUAL_DEVICE_UDN
 from .ouimeaux_device.api.service import REQUESTS_TIMEOUT
-from .util import etree_to_dict, get_ip_address, interface_addresses
+from .util import etree_to_dict, get_callback_address, interface_addresses
 
 DISCOVER_TIMEOUT = 5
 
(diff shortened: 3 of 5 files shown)


pywemo/subscribe.py
```diff
@@ -219,6 +219,16 @@ def path(self) -> str:
...
         return f'/sub/{self.service_name}'
 
+    @property
+    def callback_url(self) -> str:
+        """URL that the device will use to send push notifications."""
+        pywemo_callback_address = os.getenv('PYWEMO_CALLBACK_ADDRESS')
+        if pywemo_callback_address is not None:
+            return f'http://{pywemo_callback_address}{self.path}'
+
+        host = get_ip_address(host=self.device.host)
```
esev: Let's move the `os.getenv('PYWEMO_CALLBACK_ADDRESS')` into the `get_ip_address` method, and just use `PYWEMO_CALLBACK_ADDRESS` for the IP address only, not the port. That way this works for other callers of `get_ip_address` also.

pywemo/subscribe.py
```diff
@@ -241,8 +251,16 @@ class HTTPServer(ThreadingHTTPServer):
 
 def _start_server() -> HTTPServer | None:
     """Find a valid open port and start the HTTP server."""
-    for i in range(0, 128):
-        port = 8989 + i
+    requested_port = os.getenv('PYWEMO_HTTP_SERVER_PORT')
```
esev: > Please let me know if you think the constructor approach is more appropriate and I'm happy to go down that path.

Let's do both here. Pass the requested port in as an optional constructor arg and...

```python
def __init__(self, requested_port: int = 0) -> None:
    self.requested_port = requested_port
...


requested_port = int(os.getenv('PYWEMO_HTTP_SERVER_PORT', self.requested_port))
if requested_port:
...
```

I think it's okay to fail with a `ValueError` exception if the PYWEMO_HTTP_SERVER_PORT cannot be parsed as an `int`. That way there is some feedback/logging that what the user wanted had failed.

pywemo/subscribe.py
```diff
@@ -241,8 +251,16 @@ class HTTPServer(ThreadingHTTPServer):
...
+    if requested_port is not None and requested_port.isdigit():
+        start_port = int(requested_port)
+        ports_to_check = 1
+    else:
+        start_port = 8989
+        ports_to_check = 128
+
+    for i in range(0, ports_to_check):
```
esev, line 257: It was unlikely for this to fail before. Under the new logic there could be only 1 port that is attempted. Let's modify this a bit to give feedback in the form of a raised exception.

```python
    for i in range(0, ports_to_check):
        port = start_port + i
        try:
            return HTTPServer(('', port), RequestHandler)
        except OSError as error:
            last_error = error
    raise last_error
```

esev (changes requested):
I like this a lot. It helps Docker users who are unable to configure their networks. Thank you.

Just a few minor tweaks.

esev:
There are alternatives to using Docker `host` networking.
https://github.com/pywemo/pywemo/wiki/Networking#tips-for-running-on-the-same-lan

I definitely appreciate this PR though. There are some situations where the user won't have control over the Docker networks, and this helps in that situation. Feel free to update the 2nd bullet point about `Subscriptions` on the [wiki](https://github.com/pywemo/pywemo/wiki/Networking#features-that-require-pywemo-to-be-on-the-same-lan) after this PR is submitted.

pywemo/subscribe.py
pumbaasdad: Sounds good, but if we do that, I think I'm going to have to add a third environment variable for the host callback port.

`PYWEMO_HTTP_SERVER_PORT` controls the internal port, so if I start a container with `-p 123:456`, we need a way to pass the host port to the device.

pumbaasdad:
> There are alternatives to using Docker `host` networking. https://github.com/pywemo/pywemo/wiki/Networking#tips-for-running-on-the-same-lan
> 
> I definitely appreciate this PR though. There are some situations where the user won't have control over the Docker networks, and this helps in that situation. Feel free to update the 2nd bullet point about `Subscriptions` on the [wiki](https://github.com/pywemo/pywemo/wiki/Networking#features-that-require-pywemo-to-be-on-the-same-lan) after this PR is submitted.

Thanks for the pointer.  I completely missed the wiki when looking for a solution.  I have plans to set up an IoT VLAN at some point in the future, so having that information will definitely come in handy.

pywemo/subscribe.py
pumbaasdad: > Let's do both here.

With the commit I just pushed, the precedence will be:

 * options passed to the constructor
 * environment variables
 * auto-detected values

This order makes sense to me.  If an application provides a way to configure these values, its probably better to take those values then value from the environment.  An argument could probably also be made to reverse the order, so I'm open to changing it as well.

pywemo/subscribe.py
esev: Ah! I see. That's an edge case that I didn't account for. Thanks for clarifying.

The callback address & port also needs to match over in [ssdp.py](https://github.com/pywemo/pywemo/blob/44184ace8e7ba9a75c44337a0b699179221c7215/pywemo/ssdp.py#L478). WDYT about renaming `get_ip_addr` to `get_callback_address`, like this:

```python
def get_callback_address(host: str, port: int) -> str | None:
    """Return the IP address & port to which devices can send push notifications."""
    pywemo_callback_address = os.getenv('PYWEMO_CALLBACK_ADDRESS')
    if pywemo_callback_address is not None:
        return pywemo_callback_address

    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        sock.connect((host, 9))
        return f'{sock.getsockname()[0]}:{port}'
    except OSError:
        return None
    finally:
        del sock
```

Then, in ssdp.py, I suggest the following changes.

1. In [SSDP_NOTIFY](https://github.com/pywemo/pywemo/blob/44184ace8e7ba9a75c44337a0b699179221c7215/pywemo/ssdp.py#L46) & [SSDP_REPLY](https://github.com/pywemo/pywemo/blob/44184ace8e7ba9a75c44337a0b699179221c7215/pywemo/ssdp.py#L35) change the `http://%s:%d/setup.xml` to `http://%s/setup.xml`
2. Update `send_notify` to call get_callback_address [here](https://github.com/pywemo/pywemo/blob/44184ace8e7ba9a75c44337a0b699179221c7215/pywemo/ssdp.py#L420). Calling it with `get_callback_address(addr, self.callback_port)` will work.
3. Update `respond_to_discovery` to use get_callback_address [here](https://github.com/pywemo/pywemo/blob/44184ace8e7ba9a75c44337a0b699179221c7215/pywemo/ssdp.py#L478). Calling it with `get_callback_address(sock_addr[0], self.callback_port)` will work.

pywemo/subscribe.py
esev: Sounds good to me. Thank you.

pywemo/subscribe.py
pumbaasdad: It took me a while to wrap my mind around this, but I think I've got it working.

Based on this change, I think homeassistant needs to be updated [here](https://github.com/home-assistant/core/blob/1a43f107c492c6d8e531904188a532af1b251cd0/homeassistant/components/wemo/__init__.py#L110) to use `registry.callback_port` instead of `registry.port`.

The other option is to update `SubscriptionRegistry::port` to do what `callback_port` is doing (and then just get rid of callback_port).  I didn't go down this path as it seems like a potentially breaking change - something that relies on `port` being the local port instead of being the port used for callbacks.

pywemo/util.py
```diff
@@ -69,11 +70,25 @@ def interface_addresses() -> list[str]:
     return addresses
 
 
-def get_ip_address(host: str = '1.2.3.4') -> str | None:
-    """Return IP from hostname or IP."""
+def get_callback_address(
+    device_host: str,
+    callback_port: int,
+    callback_address: str | None = None,
```
esev: Let's drop the `callback_address` parameter. I think plumbing it through as an argument is adding complexity. Let's go with returning PYWEMO_CALLBACK_ADDRESS if it exists, or the socket/connect/getsockname logic if it is unset.

pywemo/subscribe.py
```diff
@@ -440,13 +467,28 @@ def sleep(secs: float) -> None:
...
+        self._callback_port: int | None = callback_port
 
     @property
     def port(self) -> int:
         """Return the port that the http server is listening on."""
         assert self._httpd
         return self._httpd.server_address[1]
 
+    @property
+    def callback_port(self) -> int:
```
esev: I think we don't need `callback_port` anymore. Clients (the Subscription & ssdp.DiscoveryResponder) can use `.port`.

`get_callback_address` can translate to the correct IP & port for the callback when `PYWEMO_CALLBACK_ADDRESS` is supplied.

That way
* PYWEMO_CALLBACK_PORT is used internally for the local port
* PYWEMO_CALLBACK_ADDRESS provides the IP and port to the WeMo's callback parameter.

esev (changes requested):

pywemo/subscribe.py
esev: Sorry for not being clearer initially. I had only intended to request that the `PYWEMO_HTTP_SERVER_PORT` could be passed in as a constructor arg. I was thinking `PYWEMO_CALLBACK_ADDRESS` would require more extensive changes.

pumbaasdad: No worries, I don't mind iterating when I'm getting prompt feedback :).

It's a new codebase for me, so I was gathering context as I went.

esev:
This looks great. Thank you!

esev:
The new release ([0.9.0](https://github.com/pywemo/pywemo/releases/tag/0.9.0)) is up on pypi now.
https://pypi.org/project/pywemo/

Do you want to submit a version bump PR for Home Assistant? If not, I can take care of it later tonight.

pumbaasdad:
I won't be able to get to it until next weekend, so I'll leave it to you.

esev:
Just got it merged into the Home Assistant `dev` branch. I'm hopeful they'll release it in one of the bug-fix releases later this month.

onlyoneme:
Does it ever try co connect to itself using callback address? Actually it's not enough to use just simple dnat for Wemos on VLAN and dockerized HA on main LAN because of check in get_callback_address(). I guess kind of loopback/hairpin nat must be also defined to get it working that way.

PR #375, opened 2023-05-10
Use pre-commit for code checks in build.sh

esev:
## Description:

This merges the pre-commit checks with the code health checks that were in build.sh.

This option was mentioned in #250:

> We could replace a bit of the build.sh file with just pre-commit run --all-files, but I personally prefer them all called separately in the build.sh files myself. Plus then it would only check with black and isort, not actually format.

I've modified the black & isort hooks so they can actually do formatting. A formatting change should fail the pre-commit hook but will still leave the files in their formatted/modified state. In my testing any formatting change did result in the hook failing as expected, but I added a `Validate no code changes were made` double-check in build.yml just to be extra sure.

I kind of like this approach because it removes some of the duplication between the two files. I'm planning to add[1] additional checks in the future and it'd be handy to only need to specify them in one place.

Plus as part of implementing this change, a few corner cases were caught with files that were not excluded from pre-commit. Those corner cases otherwise might not have been caught without running pre-commit with the `--all-files` option.

@Spectre5 I'm curious about your thoughts here.

[1] https://github.com/pywemo/pywemo/compare/93e26761f5ad185add635cf83b8e0b4f831ac7ba...esev:pywemo:009790dcf85aadc5ded254bbd2f32e8b3f23de4d

Related issue: #385

## Checklist:


Spectre5:
Overall it looks pretty good to me.  Unfortunately, the Python packaging ecosystem is still a mess - I wish we had rust's cargo for Python!  I've personally moved away from Poetry and now use [hatch](https://github.com/pypa/hatch), though [pdm](https://github.com/pdm-project/pdm) is another good option, particularly if you still want lock files as hatch does not yet implement them.

Another option is to consider using [nox](https://github.com/wntrblm/nox) instead of the few shell scripts.  If you're familiar with tox, it is similar but uses Python instead of an ini file for the input.  Once setup, you can just type `nox` on the command line to run all of the various tasks like format, lint, test, docs, etc.  You can then only install the test dependencies when testing, the lint dependencies when linting, etc.  With this combination I've moved away from pre-commit.  I just run `nox` before I commit, or else find out the mistake in the CI job (which can also just call nox sessions).

I would also suggest replacing flake8 with [ruff](https://github.com/charliermarsh/ruff).  Ruff is basically flake8 plus a bunch of common flake8 plugins written in rust.  It is insanely fast.  It is also more powerful as it can (optionally) auto-fix many of the lint rules it implements as well.  I've switched most of my repositories over to ruff now and I've found zero downsides.  It can also replace isort as well, and apparently has plans to eventually auto-format too (and thus replace black).  The ruff configuration is done in pyproject.toml, and mypy can also be configured in pyproject.toml now too, thus with this change we could also remove the setup.cfg file.

This is a lot of changes and if you're not familiar with these tools then it can be somewhat daunting.  If you're curious what it'd look like, I can whip up a PR to check it out...

esev:
> Overall it looks pretty good to me. Unfortunately, the Python packaging ecosystem is still a mess - I wish we had rust's cargo for Python! I've personally moved away from Poetry and now use [hatch](https://github.com/pypa/hatch), though [pdm](https://github.com/pdm-project/pdm) is another good option, particularly if you still want lock files as hatch does not yet implement them.

Initially I found the lock files a bit confusing. But now that I'm more used to them I definitely appreciate their purpose. Right now, for example, vcrpy has an incompatibility with the latest version of urllib3. Without the lock I think the pyWeMo CI would be broken right now due to all the vcrpy tests breaking.

The downside, as I found a few weeks ago, is that pinning packages that are too old can also cause breakages ([example](https://github.com/pywemo/pywemo/actions/runs/4841572127/jobs/8627923068#step:4:215)). I'm starting to wonder if a monthly dependabot auto-update would be worthwhile. But I see dependabot might not support pdm. I think for now Poetry might still be a good fit.

> Another option is to consider using [nox](https://github.com/wntrblm/nox) instead of the few shell scripts. If you're familiar with tox, it is similar but uses Python instead of an ini file for the input. Once setup, you can just type `nox` on the command line to run all of the various tasks like format, lint, test, docs, etc. You can then only install the test dependencies when testing, the lint dependencies when linting, etc. With this combination I've moved away from pre-commit. I just run `nox` before I commit, or else find out the mistake in the CI job (which can also just call nox sessions).

This looks very interesting to me! Do you know if there is something similar to [pre-commit.ci](https://pre-commit.ci) for nox?  I had seen the pairing of dependabot + pre-commit.ci as a way to keep locked dependencies periodically updated without much manual effort. That's where I was heading with these changes: https://github.com/pywemo/pywemo/compare/93e26761f5ad185add635cf83b8e0b4f831ac7ba...esev:pywemo:009790dcf85aadc5ded254bbd2f32e8b3f23de4d

But maybe there is a better way to automate periodic locked dependency updates. Is there anything that you're aware of that'll auto send PRs periodically?

> I would also suggest replacing flake8 with [ruff](https://github.com/charliermarsh/ruff). Ruff is basically flake8 plus a bunch of common flake8 plugins written in rust. It is insanely fast. It is also more powerful as it can (optionally) auto-fix many of the lint rules it implements as well. I've switched most of my repositories over to ruff now and I've found zero downsides. It can also replace isort as well, and apparently has plans to eventually auto-format too (and thus replace black). The ruff configuration is done in pyproject.toml, and mypy can also be configured in pyproject.toml now too, thus with this change we could also remove the setup.cfg file.

This sounds awesome! If you have some time, could you send a PR for this?

> This is a lot of changes and if you're not familiar with these tools then it can be somewhat daunting. If you're curious what it'd look like, I can whip up a PR to check it out...

I'm super out-of-date on all of these tools. I'm definitely interested in both nox & ruff. And if there are good solutions for keeping pinned/locked dependencies updated I'm interested in anything there too. Any examples or help you can provide is always appreciated!

BTW: I think this fork may have been accidentally created. Do you think we can remove it? https://github.com/pywemo/nox

Spectre5:
I'll follow up again soon!  As for that fork, we can certainly remove that!  Did I accidentally do that?  Don't recall that ha.

esev:
Thank you! I added another example of what I'm hoping to automate a bit in #376 

I'm thinking about enabling both the dependabot and pre-commit.ci (lite) bots. That way when dependabot sends a PR to update a dependency, pre-commit.ci can add a second commit to the PR with all the dependent files that need updating. In theory the bots would do all the necessary work and update all the right files. And then the CI would test the modifications. It'd make it very easy for me to review all the necessary changes on one PR.

Maybe there is a better solution for this though. I'd like to keep the build dependencies locked, but have some automation that updates the locked versions periodically and also updates all the generated files too. I'm definitely open to suggestions (including custom CI steps).

.pre-commit-config.yaml
```diff
@@ -32,6 +31,11 @@ repos:
         entry: poetry run flake8
         require_serial: true
         types: [python]
+        # These are auto-generated by generateDS and will not pass flake8.
+        # Normally these would be excluded by the 'exclude' filter in
+        # setup.cfg, but this is ignored when files are passed on the command
+        # line. See https://github.com/pre-commit/pre-commit-hooks/issues/112
+        exclude: pywemo/ouimeaux_device/api/xsd/
```
esev: This exclude can likely be removed after #376. I added some logic to `disable_code_analyzers` with file level comments.

esev:
@Spectre5 okay with you if I merge this?

I don't intend this to preclude us using `nox` in the future. Just, since we're already using `pre-commit`, this was the least amount of friction for now to implement #385.

Spectre5:
Yup, I just looked through it again and I'm happy.  I'll still do a PR soon regarding a few other ideas that we can test!  I'll try to get to it that this weekend.

PR #390, opened 2023-05-19
Use poetry-pre-commit-plugin to install git pre-commit hooks

esev:
## Description:

[poetry-pre-commit-plugin](https://github.com/vstrimaitis/poetry-pre-commit-plugin) will install the `pre-commit` hooks when `poetry install` is run. I'd like to use this instead of running `pre-commit install` in `build.sh` as this will work on Windows also.

## Checklist:

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1496,6 +1496,21 @@ poetry = ">=1.3.0,<2.0.0"
 poetry-core = ">=1.3.0,<2.0.0"
 
 [[package]]
+name = "poetry-pre-commit-plugin"
+version = "0.1.2"
+description = "Poetry plugin for automatically installing pre-commit hook when it is added to a project"
+category = "dev"
+optional = false
+python-versions = ">=3.7,<4.0"
+files = [
+    {file = "poetry-pre-commit-plugin-0.1.2.tar.gz", hash = "sha256:7afd327a45e14b5e17dc7fc7747f1b6f34aae421b8226368d4ca98772e823e52"},
+    {file = "poetry_pre_commit_plugin-0.1.2-py3-none-any.whl", hash = "sha256:a245945c776147df7108071e3a3699874c2bf9859dd613f6017ae1c8857fd4a4"},
+]
+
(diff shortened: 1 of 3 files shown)


esev:
Closing this. AFAICT Poetry does not load any plugins on the first run of `poetry install`.

Spectre5:
Really?  So it won't load any plugins on the first run of install, but then will on subsequent ones?  That seems like unexpected behavior to me

esev:
It doesn't look like Poetry even loads the modules on the first run. I added just an open('file', 'w+').write('test') at the top level in my plugin and that never ran. Bummer.

Spectre5:
It did run on calling install again though?

esev:
It did, yes. But I was hoping to make this a bit more seamless. If another command needs to be run, I'm thinking why limit ourselves to working within the confines of the plugin API. It's likely simpler to do something in bash as you suggested initially in #413.

```bash
echo "===Installing pre-commit hooks==="
pre-commit install
sed -ie \
    "s#INSTALL_PYTHON=.*#INSTALL_PYTHON=$(pwd)/scripts/pre-commit.sh#" \
    .git/hooks/pre-commit
```

Spectre5:
It seems really bizarre that the plugin system would not work the first time install is called.  Surely any plugin someone would perform at the end of install they would want to be performed the first time too.  Seems like a bug, but I have no experience with poetry plugins...

esev:
I agree, it does seem like a bug. I wonder if the plugins are meant to be installed on the host and not as part of a project. That's probably why the poetry-pre-commit-plugin [installation instructions](https://github.com/vstrimaitis/poetry-pre-commit-plugin#installation) suggest using "poetry **self** add ..."

Spectre5:
Ah, right, that's exactly it.  You're "supposed" to install it into the same environment that poetry is installed in (so poetry self or pipx inject if using pipx, like me).  So the first time you run it, the plugin gets installed.  Then in the subsequent runs, the plugin is installed...

Issue #385, opened 2023-05-18
Automated dependency updates


esev:
I propose we turn on [Dependabot](https://docs.github.com/en/code-security/dependabot) for pyWeMo.

When working on #360 I encountered some out-of-date dependencies that prevented the build CI workflow from working ([details](https://github.com/pywemo/pywemo/actions/runs/4841572127/jobs/8627923068#step:4:215)). It looks like the GitHub runner had a newer and incompatible version of `virtualenv` that did not work with pyWeMo's locked dependencies. It had been at least a year since the dependencies had been updated.

## Let's enable Dependabot

Dependabot should prevent this from happening again. If enabled, Dependabot would periodically create PRs to update the dependencies in `pyproject.toml` & `poetry.lock`. And pyWeMo's Build CI workflow would test the dependabot PRs to ensure there were no breakages.

There are a few things that need to be considered first though. We need to make sure that any side-effects of updating the dependencies are also handled. These include:

1. Regenerating the Python modules in `pywemo/ouimeaux_device/api/xsd` whenever the version of `generateDS` is updated.
2. Ensuring that the build/test/publish process uses a consistent version of Poetry. Ideally Dependabot will send PRs to update the Poetry version and when we install Poetry (`pip install poetry`) the version that gets installed should match the updated version.
3. New versions of formatters like `black` and `isort` can produce different results. We need to ensure Dependabot can adjust the existing files when that happens.

## Handle dependency side-effects with pre-commit

pyWeMo currently uses `pre-commit` to ensure code is formatted correctly and linted before creating a commit. That takes care of case 3 above. We can also extend our `pre-commit` configuration so it can handle 1 & 2 also. I've addressed these in #376 & #382. This way `pre-commit` can automatically adjust for the side-effects of dependency updates. But it currently only works for git commits that are run locally. Dependabot will not run `pre-commit` before sending a PR.

To have `pre-commit` apply to Dependabot PRs I propose we make a few changes to pyWeMo's Build CI workflow.

1. Run `pre-commit` as part of the Build CI workflow. This takes care of updating any files that needed to be modified as a side-effect of updating a dependency.
2. Save any changes that `pre-commit` makes during the CI workflow.
3. Apply those changes as a commit on the Dependabot PR.

[...]

PR #395, opened 2023-05-20
Push modified files as a commit on a PR

esev:
## Description:

This is intended to be used in conjunction with Dependabot as described in #385.

When dependencies are updated, some side-effects also need to happen. This can include reformatting the code, or re-generating some auto-generated files. #375 added the necessary Build steps to save the side-effect diffs. This change is intended to push the needed side-effects onto PRs for dependency updates. Changes will be authored by `pyWeMo[bot]`.

I've intentionally limited the scope of this workflow to just Dependabot PRs and just to modified source files (no adds/deletes/renames). The workflow has access to an App (pyWeMo) with commit access permissions to pyWeMo. I'd like to be cautious for now on using those permissions.

**Related issue (if applicable):** #385

## Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,13 +34,15 @@ jobs:
         if: always()
         id: modified
         run: |
-          mkdir -p ./.cache  # This directory is in .gitignore
-          git diff "${GITHUB_SHA}" | tee ./.cache/build-changes.patch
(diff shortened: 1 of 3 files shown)


esev:
A good example of this in action: https://github.com/esev/pywemo/pull/13

PR #411, opened 2023-05-22
Update mypy config

Spectre5:
## Description:
There was still old/stale mypy configuration options in the setup.cfg file (mypy options in pyproject.toml take precedence).  I've removed those and merged a couple into pyproject.toml.  Most of the options enabled are part of the "strict = true" rule, so I switched to that (which then enables 3 extra rules that were not already enabled (disallow_any_generics, no_implicit_reexport, and strict_concatenate).  I remove a rule or two that I believe no longer exists.

I switched  to make the two cases of ignore_missing_imports explicit via an override section.  Note that lxml has two different types packages that could be added to the dev dependencies and then it could be removed from this override.  I can try them both out later to see what other changes that might require and which of the two are better.

@esev, note that the pre-commit process still isn't exactly "fool-proof", as you probably already suspect.  I recently updated to Poetry 1.5.0, which is new.  When I tried to commit my code, that caused poetry to update the lock file (added a catgeory="dev" section to those dependencies) and bootstrap file.  Even though the bootstrap file locks Poetry to 1.4.2, when you run pre-commit it just calls "poetry run ...." and thus uses your system-installed version of poetry.  This assumes, of course, that the virtual environment is not already active (mine generally isn't) - though I know @esev works with it active while developing and thus has probably not seen this issue.  So I committed with --no-verify to prevent that update to the lock file.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,22 +109,17 @@ src_paths = ["pywemo", "tests"]
 skip_glob = "*/xsd/*"
 
 [tool.mypy]
-show_error_codes = true
-ignore_missing_imports = true
-strict_equality = true
+cache_dir = ".cache/mypy/"
+strict = true
 warn_incomplete_stub = true
-warn_redundant_casts = true
-warn_unused_configs = true
-warn_unused_ignores = true
-check_untyped_defs = true
-disallow_incomplete_defs = true
-disallow_subclassing_any = true
-disallow_untyped_calls = true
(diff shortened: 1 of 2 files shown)


esev (approved):
Thanks for this! Nice to see `strict = true` replacing all those other settings.

I filed #413 for the issue with the Poetry version mismatch. I'll find a better way.

PR #418, opened 2023-05-23
Always run pre-commit inside the venv

### LINKED ISSUE #413: Poetry version mismatch causes pre-commit to fail.
The pre-commit checks can fail when the Poetry version on the host is different from the Poetry version used by pyWeMo. More details on #411.

I believe this is happening due to the `poetry-lock-check` added in #382.

esev:
## Description:

The pre-commit git hook (.git/hooks/pre-commit) currently runs [pre-commit](https://github.com/pre-commit/pre-commit) outside of a venv. This PR modifies pre-commit's git hook run pre-commit inside the venv.

This is based on https://github.com/cjolowicz/nox-poetry/blob/5772b66ebff8d5a3351a08ed402d3d31e48be5f8/noxfile.py#L70

**Related issue (if applicable):** fixes #413

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
       - id: generateDS
         name: generateDS
         language: system
-        entry: poetry run python scripts/generate_ds.py
+        entry: python scripts/generate_ds.py
         args:
           - pywemo/ouimeaux_device/api/xsd/device.xsd
           - pywemo/ouimeaux_device/api/xsd/service.xsd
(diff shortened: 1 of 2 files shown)


Spectre5:
So with this change, I think you can also speed up pre-commit by removing "poetry run" from all instances within `.pre-commit-config.yaml`.  Because if pre-commit now activates the virtual environment, then we can just call black, flake8, isort, etc directly and get the correct one.

esev:
Ah, yes. Good call.

PR #420, opened 2023-05-26
Handle invalid data received from device

esev:
## Description:

Apply some error checking on Device responses to catch cases where invalid data is present. Add [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) tests to improve coverage for these edge cases.

## Checklist:
  - [x] The code change is tested and works locally.
  - [x] There is no commented out code in this PR.
  - [x] I have read and agree to the [CONTRIBUTING document](
        https://github.com/pywemo/pywemo/blob/master/CONTRIBUTING.md).

PR #449, opened 2023-06-07
Use ClusterFuzzLite for fuzzing the NOTIFY http endpoint 

esev:
## Description:

Use ClusterFuzzLight/OSS-Fuzz for a more in-depth fuzzing of the NOTIFY http endpoint. This also includes support for Atheris to direct the fuzzing using code coverage. All of the recent PRs to fix validation issues have been surfaced by these additional tests: #444, #443, #434, #432, #421, and #420. I think there is value in running this fuzzer quickly on PRs and for longer during scheduled batch runs (once per day for 1 hour).

One quirk: Atheris isn't supported on Windows, so I haven't included it as a dev dependency in pyproject.toml and I've added a small `tests/atheris_stub.py` file to keep pytest working even when Atheris is not installed. This requires that Atheris be installed manually to run the fuzzer manually.

```bash
. .venv/bin/activate
pip install atheris
python tests/test_registry_notify_fuzz.py
```

These workflows have been running successfully in my fork for the past couple weeks: https://github.com/esev/pywemo/actions/workflows/fuzz.yml

## Checklist:

### DIFF
--- a/.clusterfuzzlite/Dockerfile
+++ b/.clusterfuzzlite/Dockerfile
@@ -1,0 +1,6 @@
+FROM gcr.io/oss-fuzz-base/base-builder-python:v1
+RUN apt-get update && \
+    apt-get install -y make autoconf automake libtool pip
+COPY . $SRC/pywemo
+WORKDIR pywemo
+COPY .clusterfuzzlite/build.sh $SRC/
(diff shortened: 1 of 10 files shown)


Spectre5:
How was the requirements file created?  I think it needs to be updated...

PR #482, opened 2023-06-21
Fix missing SSDP headers

esev:
## Description:

Fix the SSDP headers to match more closely with what a WeMo device would send. This is based on the headers in https://sourceforge.net/p/pupnp/code/ci/release-1.8.0/tree/upnp/src/ssdp/ssdp_device.c#l317

## Checklist:

### DIFF
--- a/pywemo/ssdp.py
+++ b/pywemo/ssdp.py
@@ -6,7 +6,10 @@ import re
 import select
 import socket
 import threading
+import time
+import uuid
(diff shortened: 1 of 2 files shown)


pywemo/ssdp.py
```diff
@@ -28,9 +31,13 @@
```
Stefan41199, line 32: - [ ]

Stefan41199 (approved):

PR #632, opened 2023-10-22
Support for DimmerV2 transition times

### LINKED ISSUE #254: Add fader support for Dimmers
WeMo dimmers support a fade transition time for transitioning between brightness levels. This is supported for the Bridge device in pywemo, but not for Dimmers. It would be nice to have this supported in pywemo for Dimmers also.

`SetBinaryState(BinaryState, Duration, EndAction, brightness, fader, UDN, CountdownEndTime, deviceCurrentTime)`

The `fader` parameter appears to be the key to making this happen. The fader parameter requires 5 colon separated values (ex: `300:-1:1:0:33`). Here's a starting point on what these parameters might mean (this is my best guess at the parameters thus far):

1. Fade time in seconds
2. `-1`=fader not active; otherwise this appears to be a timestamp for when the fade started. Setting it to `0` appears to use the current timestamp.
3. Still trying to work out this one. 
   - Setting it to `0` starts a timer but doesn't cause any fade. At the end of the timer the Dimmer turns off and the brightness level remains at its original state.
   - Setting this to `1` causes the Dimmer to turn on and fade from the current brightness to the target brightness.
4. Floating point value representing the number of brightness levels to change per second. This is a computed value, setting this value seems to have no effect.
5. Target brightness - fade to this brightness level. 0 turns the device off after fading.

There is a green light on the device that remains active throughout the fade transition. If the device turns off after the fade, then this green light also turns off. If the device is on after the fade, this green light remains on too. The green light can be turned off by setting parameter 2 to `-1`. This seems to cancel the fade, leaving the brightness at its current setting (halting the fade at the current transitional brightness if it was still in progress).

If the Device is off, the previous brightness will be restored before activating the fader. If the light was at 100% brightness previously it will turn on to that brightness and then fade to the target brightness. This likely isn't desirable. If the device is off, it's probably best to set the brightness to 1% before starting a fade.

Touloused:
## Description:
Add support for transition times ("fader" options in Wemo parlance) for `on`, `off`, `toggle`, and `set_brightness` methods on `DimmerV2` .

Support is limited to `DimmerV2` models (WDS060). That's the only model I personally have access to, ~~and from my limited understanding, the only model that supports this (?) other than the `Bridge`.~~  _It seems it might be possible to get this working for older dimmer models if anybody wants to help._

**Related issue (if applicable):** implements #254

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -1,6 +1,8 @@
 """Representation of a WeMo Dimmer device."""
 from __future__ import annotations
 
+import time
+
 from .api.long_press import LongPressMixin
 from .api.service import RequiredService
 from .switch import Switch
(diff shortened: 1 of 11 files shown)


Touloused:
<H1>🤔</H1> 

I actually see the "fader" parameter for `SetBinaryState` showing up for older dimmers also (see [tests.ouimeaux_device.test_dimmer/WeMo_WW_2.00.11453.PVT-OWRT-Dimmer.yaml](https://github.com/pywemo/pywemo/blob/main/tests/vcr/tests.ouimeaux_device.test_dimmer/WeMo_WW_2.00.11453.PVT-OWRT-Dimmer.yaml#L255)).

Anyone want to help to see if we can get this to work for them?
(As mentioned in the OP, I only have physical access to the newer dimmer model)

esev:
> Anyone want to help to see if we can get this to work for them?

I have a few of these and can give it a test after work tonight.

Spectre5:
@Touloused can you rebase this?

Spectre5:
Also, can you provide the firmware that your device has and the SKU of the device:
https://www.belkin.com/uk/support-article/?articleNum=335419

Have you tried the `setup` method to connect this device to an AP? If so, does it work?  With Belkin support ending soon, I'd like to document which devices and firmware are known to work with `setup`, which allows pywemo to connect a device to an AP without Belkin nor the cloud.

If you have any other WeMo devices, I'd also like to connect the information on them.  See https://github.com/pywemo/pywemo/issues/773.

Spectre5:
> # 🤔
> 
> I actually see the "fader" parameter for `SetBinaryState` showing up for older dimmers also (see [tests.ouimeaux_device.test_dimmer/WeMo_WW_2.00.11453.PVT-OWRT-Dimmer.yaml](https://github.com/pywemo/pywemo/blob/main/tests/vcr/tests.ouimeaux_device.test_dimmer/WeMo_WW_2.00.11453.PVT-OWRT-Dimmer.yaml#L255)).
> 
> Anyone want to help to see if we can get this to work for them? (As mentioned in the OP, I only have physical access to the newer dimmer model)

@Touloused I have one of the older dimmers and am interested in trying this out.  It looks like the same code might work for the older dimmer too, so perhaps we can just move your new code into the parent Dimmer class?

PR #633, opened 2023-10-23
Fixed rare incorrect startup brightness value for Dimmers

Touloused:
## Description:

Stumbled upon and fixed this random minor issue while working in the repo. Unlikely that it appears as a problem downstream, just weird behaviour when writing local tests.

The bug was happening if the first operation performed on a `Dimmer` was to set its basic state value (ie. `on`, `off`, or `toggle`), followed by a request for `get_brightness(force_update=False)`. This would invoke the parent `Switch` class' 
`get_state(force_update=False)` which would early out because it already has a valid basic state. `Dimmer` would then naively look in the `basic_state_params` to try to find its "brightness" value, which it would not find, and then humbly fallback to setting its internal `_brightness` to zero (which is actually not a valid value for this parameter).

Fixed by calling an explicit `Switch.get_state(force_update=True)` in the appropriate circumstance.

**Related issue (if applicable):** fixes _an unlogged bug_

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/dimmer.py
+++ b/pywemo/ouimeaux_device/dimmer.py
@@ -38,6 +38,10 @@ class Dimmer(Switch):
 
     def get_state(self, force_update: bool = False) -> int:
         """Update the state & brightness for the Dimmer."""
+        force_update = force_update or (
+            self._brightness is None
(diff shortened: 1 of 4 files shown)


esev (approved):
This looks great to me. Thank you for the fix!

PR #700, opened 2024-02-04
Use ruff

Spectre5:
## Description:
This is an initial proof of concept on replacing isort, black, and flake8 with ruff.  Note that I've selected ALL rules for ruff and then selectively disabled some, which has something like 700 rules (way more than flake8 alone).  The reason I like to do this is so that I don't miss future rules that get added.  If you don't like some future rule, then it can be disabled.  Ruff can also autofix some of the rule violations it finds, which is a really nice feature too.

At the moment, the build doesn't work.  After a few tweaks to the config and a few files, there are roughly 600 violations still, though about 200 can be automatically fixed by ruff (not done at the moment).  Take a look @esev and see what you think.  If we like it (ruff is built in rust and thus is WAY faster, basically instant, even with so many rules), then I can look at finishing it up.  But we'd maybe want some discussion on the remaining rules to see if we want to disable some or modify the code to "fix" them all.

(note that one rule includes a check for commented out code - so that could help remove number 3 in the checklist below)

## Checklist:


Spectre5:
Closed in favor of #762

PR #762, opened 2025-10-21
Use Ruff

Spectre5:
## Description:
This work replaces black, isort, and flake8 with ruff.  Note that I've selected ALL rules for ruff and then selectively disabled some.  The reason I like to do this is so that I don't miss future rules that get added. If you don't like some future rule, then it can be disabled. Ruff can also autofix some of the rule violations it finds, which is a really nice feature too.

NOTE: this is a re-creation of #700, where ruff is updated to a newer version and MANY ruff codes are disabled to avoid disrupting the code base too much.  In the future, each of the ignored codes in ruff.lint should be investigated and either fixed in the code, moved to the "permanent" ignore list (above the TODO), selectively disabled in the code with noqa comments, or ignored per file using [tool.ruff.lint.per-file-ignores].

This re-creation should build and pass the CI checks (it does locally anyway) due to all of the additional (temporarily) ignored rules.

## Checklist:


Spectre5:
I'm not familiar with ClusterFuzzLite or what the exact purpose is in this code base, so perhaps @esev can take a look at the CI failure.  It seems to be a complaint from pip that the pip installation is not pinned with ==, but it seems that it is.  More details below.

CI Error:
```
+ pip install --require-hashes --no-deps --only-binary :all: -c .clusterfuzzlite/requirements.txt pip
Collecting pip
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    pip from https://files.pythonhosted.org/packages/b7/3f/945ef7ab14dc4f9d7f40288d2df998d1837ee0888ec3659c813487572faa/pip-25.2-py3-none-any.whl
```

But the pip constaint `.clusterfuzzlite/requirements.txt` includes
```
pip==25.2 ; python_full_version >= "3.8.1" and python_version < "3.12" \
    --hash=sha256:578283f006390f85bb6282dffb876454593d637f5d1be494b5202ce4877e71f2 \
    --hash=sha256:6d67a2b4e7f14d8b31b8b52648866fa717f45a1eb70e83002f4331d07e953717
```

Did you have similar issues when first implementing this?

PR #814, opened 2026-01-31
2.0.0 release

esev:
## Description:

Update the PyWeMo version to 2.0.0.


Spectre5:
@esev, look goods to me.  I've been wanting to find the time to complete #632 and #772, but not sure exactly when I'll get to that and so we shouldn't wait for it.

Otherwise, pay attention to what the CI/CD does after this is merged (the docs and deploy ones).

Issue #804, opened 2026-01-30
F7C030 Not Connecting


kilpatrick36:
> # F7C030 (setup does work)
> 
> ```python
>>> type(device)
> <class 'pywemo.ouimeaux_device.lightswitch.LightSwitchLongPress'>
>>> device.firmware_version
> 'WeMo_WW_2.00.11408.PVT-OWRT-LS'
>>> pprint(device._config_any)
> {'binaryState': '0',
>  'firmwareVersion': 'WeMo_WW_2.00.11408.PVT-OWRT-LS',
>  'iconVersion': '0|49153',
>  'new_algo': '1'}
> ``` 

 _Originally posted by @Spectre5 in [#773](https://github.com/pywemo/pywemo/issues/773#issuecomment-3489512051)_

**I have tried all 6 versions of the setup method and none have worked. Seems I have the newest firmware and I am a novice at python but here are the logs:** 

```
PS C:\WINDOWS\system32> python3
Python 3.13.9 (tags/v3.13.9:8183fa5, Oct 14 2025, 14:09:13) [MSC v.1944 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> import pywemo
>>> from pprint import pprint
>>> logging.basicConfig(level=logging.DEBUG)
>>> type(device)
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    type(device)
         ^^^^^^
NameError: name 'device' is not defined
>>> devices = pywemo.discover_devices(debug=True)
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 10.22.22.1:49152
DEBUG:urllib3.connectionpool:http://10.22.22.1:49152 "GET /setup.xml HTTP/1.1" 200 0
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 10.22.22.1:49152
DEBUG:urllib3.connectionpool:http://10.22.22.1:49152 "GET /setupservice.xml HTTP/1.1" 200 0
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 10.22.22.1:49152
DEBUG:urllib3.connectionpool:http://10.22.22.1:49152 "GET /timesyncservice.xml HTTP/1.1" 200 0
DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): 10.22.22.1:49152
[...]

Issue #773, opened 2025-11-05
META: pywemo is not affected by Belkin killing Wemo!  Please help document pywemo status!


Spectre5:
Belkin is killing support for Wemo after January 31, 2026:
https://www.belkin.com/uk/support-article/?articleNum=335419

The good news is that pywemo is entirely local and does not rely on Belkin nor its cloud services, thus pywemo will continue to work without issue.

This will likely lead to an increase in activity to pywemo, probably mainly through the use of Home Assistant.  Pywemo can also setup a device (connect it to a new access point).

[This project wiki page](https://github.com/pywemo/pywemo/wiki/Device-Status-Registry) has a table of product status that we should ideally complete as fully as possible, including whether the `setup` method works and which firmware versions are known to work.

Please comment in this thread with any Wemo devices you have, the SKU of the device, the firmware, and whether the pywemo setup method works.  Note that if you reset a device and the pywemo setup does not work, you can still re-connect it via the Belkin app, but **only until January 31, 2026**.  Ideally the `setup` method can be shown to work (or be updated if needed) for all devices.

Here is the information that should be posted as comments:
```python
import logging
import pywemo
from pprint import pprint
logging.basicConfig(level=logging.DEBUG)
# optionally filter out the urllib3 logging
# logging.getLogger('urllib3.connectionpool').setLevel(logging.CRITICAL)
devices = pywemo.discover_devices(debug=True)
device = devices[0]  # make sure this is the intended device

# include the next 3 outputs in your comment, see example in first comment
type(device)
device.firmware_version
pprint(device._config_any)
```

Note: the SKU unfortunately needs to be looked up on the physical device itself, the device has no API to provide this information.

To test setup, you can reset the device, connect to its local wifi (something like Wemo.AAA.BBB), discover the device as shown above, and then run the setup method:
```python
device.setup(ssid='SSID', password='PASSWORD')
```

If that fails, then read through the [setup section](https://github.com/pywemo/pywemo?tab=readme-ov-file#setup) in the README and finally try the alternative methods discussed at the end.  If one of the alternative methods work, but not the default (only passing ssid and password), then please be sure to indicate that here!

Spectre5:
# F7C030 (setup does work)

```python
>>> type(device)
<class 'pywemo.ouimeaux_device.lightswitch.LightSwitchLongPress'>
>>> device.firmware_version
'WeMo_WW_2.00.11408.PVT-OWRT-LS'
>>> pprint(device._config_any)
[...]

PR #819, opened 2026-02-01
Use encrypt method 1 with iot

### LINKED ISSUE #803: I think i installed the wrong Pywemo
I followed the instructions over here at https://github.com/TomsSmartHome/PyWemo-Commands

And the version I have needs slightly different commands.

And when I tried to use it I got a bad error like this.

How do i switch to your version or can you tell me what I am doing wrong?

The method of toggling the device on and off is working so I know I am connected right and things are working.  I just can't get the darn thing to accept the credentials. I want to have this working for a backup.

I am also wondering if what Wyze means by the APP not working anymore is that the ability to control devices will not work anymore.  Because I can put my iphone in airplane mode just connect to the device wifi only with the app and re-configure it without my phone having any internet connection. I start the app after it is connected to the device wifi and i am in airplane mode.  So I am thinking MAYBE this aspect of just setting up the devices may still work in the app.   That said I would rather use your method. 

```
 `devices[0].setup(ssid=‘Ms. Xeean’, password=‘t0no00le212’)

  File "<python-input-8>", line 1, in <module>
    devices[0].setup(ssid='Ms. Xeean, password='t0no00le212’)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jdawson\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages\Python313\site-packages\pywemo\ouimeaux_device\__init__.py", line 424, in setup
    return self._setup(*args, **kwargs)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "C:\Users\jdawson\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0\LocalCache\local-packages\Python313\site-packages\pywemo\ouimeaux_device\__init__.py", line 639, in _setup
    raise SetupException(
    ...<3 lines>...
    )
pywemo.exceptions.SetupException: Wemo device failed to connect to "Mr. Clean".  It could be a wrong password or Wemo device/firmware issue.  Please try again.
>>>
```

esev:
## Description:

There are some devices (F7C027, F7C028, F7C029, F7C030) that have `'rtos': '1'` in `_config_any` but require encrypt method 1 for performing WiFi Setup. One commonality between all these devices is that they all also have `'iot': '1'` in `_config_any`. They also have a firmware version starting with `WeMo_WW_2.00.11851.PVT-OWRT-...`. Let's use this `iot` config to identify them as ones that should use method 1 instead of method 2.

### Manual testing

-  [x] F7C027 Testing by @JosephRDawson in #803
```
{'binaryOption': '1',
'binaryState': '0',
'firmwareVersion': 'WeMo_WW_2.00.11851.PVT-OWRT-SNS',
'iconVersion': '0|49152',
'iot': '1',
'new_algo': '1',
'rtos': '1'}
```

-  [x] F7C027 Testing by @esev
```
{'binaryOption': '1',
 'binaryState': '0',
 'firmwareVersion': 'WeMo_WW_2.00.11851.PVT-OWRT-SNS',
 'iconVersion': '0|49152',
 'iot': '1',
 'new_algo': '1',
 'rtos': '1'}
```

-  [x] F7C028 Testing by @esev
```
{'binaryOption': '1',
 'binaryState': '0',
 'firmwareVersion': 'WeMo_WW_2.00.11851.PVT-OWRT-SNS',
 'iconVersion': '1|49153',
 'iot': '1',
 'new_algo': '1',
 'rtos': '1'}
```

-  [x] WSP090 Testing by @esev
```
{'binaryState': '0',
 'firmwareVersion': 'WEMO_WW_1.00.20081401.PVT-RTOS-OutdoorV1',
 'hkSetupCode': 'xxx-xx-xxx',
 'hwVersion': 'v1',
 'iconVersion': '1|49152',
 'new_algo': '1',
 'rtos': '1'}
```

### Alternative considered

There is overlap between devices with `'binaryOption': '1'` and encrypt method 1 as well. But my F7C030 & F7C031 work with encrypt method 1 and also doesn't have `'binaryOption': '1'`. `iot` seems to be the thing that is in common with the devices that currently aren't working with WiFi setup.

**Related issue (if applicable):** 
- fixes #803 

## Checklist:

### DIFF
--- a/pywemo/ouimeaux_device/__init__.py
+++ b/pywemo/ouimeaux_device/__init__.py
@@ -619,7 +619,8 @@ class Device(DeviceDescription, RequiredServicesMixin, WeMoServiceTypesMixin):
 
                 # however, this works correctly more often than the logic above
                 is_rtos = self._config_any.get("rtos", "0") == "1"
-                if is_rtos:
+                is_iot = self._config_any.get("iot", "0") == "1"
+                if is_rtos and not is_iot:
                     method = 2
                 else:
                     method = 1


esev:
Keeping as draft till we hear back about ~~#803 &~~ #804

esev:
@ENCODER99 would you be able to test this on your F7C030 with WeMo_WW_2.00.11851.PVT-OWRT-LS?

```python
device.setup(ssid='iExpress24', password='replace_with_password', _encrypt_method=1)
```

If you see `Setup status is uncertain, re-probing and checking is required.`, that's not necessarily an error. Please check to see if it actually did work.

Spectre5:
Ideally we can figure out which combination of options require method 1 vs 2 (none seem to use 3)...possible combinations are options between `rtos`, `binaryOption`, `iot`, and `new_algo`.  I saw references to `new_algo` and `binaryOption` in the android app, but haven't looked closely in the actual firmware.  Looks like we need to make a list of all devices from the META issue (#773) to see if we can find any other trends between these options, which it looks like you've started here.

Another option I had considered was to just try option 1 and if that fails, then try option 2.  I ended up using those manual override options instead to just trying both to, in part, try to find out which combinations actually require one vs the other.  But just doing both could be an option if consistent logic can't be found.

Sometimes I wish I requested in that META issue to just try all 6 combinations --- apparently at least some firmware/devices will accept method 1 or 2, such as that reported by @jessanya (which also has `iot=1` and does not work by default, but the way).

JosephRDawson:
I think coding it so you just had to provide SSID and Password and then it
would say ...

Trying option #1....
Trying option #2...
Trying option #3...

And have it continue until it takes.  Even if you have to wait like 1 min
or more between tries to allow time to see if the device is turning off its
wifi.
I think that would be the best option.  I am not sure why I am getting that
error after it works for me.  But I am very happy to see it is working.

Spectre5:
@RaddedMC, it seems that you indicated before that setup does work for your F7C027 wemo switch.  First you said it did not work, but the updated your comment later and said it did.

Did you try one of the "alternative" methods to make it work (see the setup section of the README), or did you just use `setup` with ssid and password only?

Spectre5:
Looking closer at the reports from these, all of these problem devices seem to have all 4 of `rtos`, `iot`, `binaryOption`, `new_algo` set to 1?

Perhaps looking again through past issues will provide more clues.

Spectre5:
> @ENCODER99 would you be able to test this on your F7C030 with WeMo_WW_2.00.11851.PVT-OWRT-LS?
> 
> ```python
> device.setup(ssid='iExpress24', password='replace_with_password', _encrypt_method=1)
> ```
> 
> If you see `Setup status is uncertain, re-probing and checking is required.`, that's not necessarily an error. Please check to see if it actually did work.

@ENCODER99, and also can you provide the "_config_any" details for the device that worked right away (with firmwar e WeMo_WW_2.00.10062.PVT-OWRT-LS)?  I've fairly certain that it will *not* have `iot` (nor some other keys) defined.

esev:
I'd prefer we detect the appropriate mechanism for each device. It looks like the original devices use method 1 and the V2&V3 devices use method 2. I'd be willing to switch this PR to use binaryOption if we thought that was more reliable. I chose `iot` because it seemed like only these few first gen devices were the ones that needed a fix.
 
I'm also happy to try each method. We need a better solution for when the WiFi provisioning works and the device is no longer reachable on the 10.22.22.1 IP address though.

But hopefully the UI implementations don't depend on the private underscore args that we're using to debug at the moment. Or, if we intend folks to build solutions off those args, we should remove the underscore.

JosephRDawson:
You don’t need to setup devices often so if it takes a while it not a big deal. All the best,JoeOn Feb 2, 2026, at 10:47 PM, Eric Severance ***@***.***> wrote:﻿esev left a comment (pywemo/pywemo#819)
I'd prefer we detect the appropriate mechanism for each device. It looks like the original devices use method 1 and the V2&V3 devices use method 2. I'd be willing to switch this PR to use binaryOption if we thought that was more reliable. I chose iot because it seemed like only these few first gen devices were the ones that needed a fix.
I'm also happy to try each method. We need a better solution for when the WiFi provisioning works and the device is no longer reachable on the 10.22.22.1 IP address though.
But hopefully the UI implementations don't depend on the private underscore args that we're using to debug at the moment.

Spectre5:
All examples we've seen with `rtos=1` have `binaryOption` and `iot` both undefined or both 1.  So in that case, it looks like you could use either one.  There might not be any firmware where only one of them are defined with `rtos=1`.  I think `iot` is probably the better bet though, since there are cases with `rtos` undefined but `binaryOption=1`.

Spectre5 (approved):

Spectre5:
If we still find issues after this gets merged, then we can investigate with that user based on these options.

If we can't find a logical solution at that time, then we can use this logic to set the order for trying each of the solutions.

I also wonder how many devices might work with multiple methods, and if that could be related to these variables.  At least one user did apparently test all 6 and found 2 that worked (method 1 and method 2, but both with lengths, which is odd).  Perhaps I should try testing each of my devices for all 6 combinations too...

PR #820, opened 2026-02-01
Remove the product status table (moved to wiki)

Spectre5:
## Description:
This PR removes the product status table from the README and instead points to a new wiki page, which will be easier to make quick changes and updates to.

I've also made a few other updates, like removing the "Firmware warning" section, since it seems pretty unlikely that Belkin will update it now :)

## Checklist:


esev:
Thanks for pulling this all in one place!

PR #842, opened 2026-02-23
Update README.rst

kaloz:
## Description:

- use the default WeMo setup IP everywhere
- fix setup command

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,9 +24,9 @@ In that case, you can use the device with an IP or hostname:
 .. code-block:: python
 
     >>> import pywemo
-    >>> url = pywemo.setup_url_for_address("192.168.1.192")
+    >>> url = pywemo.setup_url_for_address("10.22.22.1")
     >>> print(url)
-    http://192.168.1.192:49153/setup.xml
+    http://10.22.22.1:49153/setup.xml
     >>> device = pywemo.discovery.device_from_description(url)
     >>> print(device)
     <WeMo Maker "Hi Fi Systemline Sensor">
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -88,7 +88,7 @@ Device setup is through the ``setup`` method, which has two required arguments:
 The user must first connect to the devices locally broadcast access point, which typically starts with "WeMo.", and then discover the device there.
 Once done, pass the desired SSID and password (WPA2/AES encryption only) to the ``setup`` method to connect it to your Wi-Fi network.
 
-``device.setup(ssid='wifi', password='secret')``
+``device[0].setup(ssid='wifi', password='secret')``
```
esev, line 91: There is a pattern to the readme where the plural `devices` is a list and the singular `device` is a single instance of a device.

For example, when discussing using discovery, `devices` is used:

```python
devices = pywemo.discover_devices()
...
devices[0].toggle()
```

And when not using discovery...

```python
device = pywemo.discovery.device_from_description(url)
```

Perhaps in the first example, with discovery, it should be modified like this:

```python
>>> import pywemo
>>> devices = pywemo.discover_devices()
>>> print(devices)
[<WeMo Insight "AC Insight">]

>>> device = devices[0]
>>> device.toggle()
```

Then it'd fit better with the rest of the documentation in the readme. Could you make that change instead of using `device[0]` here?

esev (changes requested):

Spectre5:
Closing due to lack of activity.  Created #944 to add the line `device = devices[0]` as @esev suggested above.

